PR #3, opened 2016-06-11
Add nvmrc + extend gitignore

mixmix:


pfraze:
The build directory isnt produced anymore, so it's not in the .gitignore.

And, you dont have to use node 6 anymore, because of the `npm run rebuild` command.

Thanks though!

PR #5, opened 2016-06-13
downgrade electron to 1.1.3

mafintosh:
Should solve the weird replication issues. See https://github.com/electron/electron/issues/5851 for more details


pfraze:
Yep, that solves it.

Issue #56, opened 2016-08-14
Beaker Plugins. Scan 'app/node_modules', instead of using global modules?


joshuef:
I've been starting with the excellent new plugin setup, but ran into a stumbling block just now. I'm not sure if it's really a bug or just a documentation fix... so I thought I'd ask before firing off a PR.

The problem I had developing was the use of `globalModules`, when scanning for plugins.

It's [noted](https://github.com/pfrazee/beaker/blob/master/doc/authoring-plugins.md#overview):
> On load, Beaker will find all modules that fit that name-scheme under ./app/node_modules, and load them.

So I was firing in with a plugin in that folder.. but it wasn't being found. I did some digging and then remembered [this bit](https://github.com/pfrazee/beaker#plugins):

>To install a new plugin, install it globally using npm. 

I was also a bit confused as `dat` and `ipfs` were running, even though I didn't `npm i -g` those plugins.

Anywhooo. This could be as simple as addressing the protocol plugin documentation just to make it clear that things should be global, but I was wondering about the use of `global-modules`, is there a reason you didn't want to reference the `app/node_modules` folder itself and use the plugins as installed there?

I think it would be cleaner to use only the modules as specified in the `app/package.json`. [I have a fork of this](https://github.com/joshuef/beaker) utilising `app/node_modules` setup and working, I can happily PR that.

pfrazee:
Sorry, yes, the documentation is wrong. I'll update it.

Originally the plugins were located in `app/node_modules`, but that didn't work for the binary package. I'm open to alternatives.

pfrazee:
Doc is updated

joshuef:
> that didn't work for the binary package. 

For compiling on OSX? Or on other platforms? The fork has compiled successfully on my osx el cap, but I don't have the facilities to check windows or linux at the mo.

pfrazee:
There are two issues

 1. Electron binaries are typically packed into the .asar archive format, which means you cant npm install into the app/node_modules. [This page explains why that's done](http://electron.atom.io/docs/tutorial/application-packaging/), but the only reason that I care about is the Windows paths. If we find the paths arent an issue in Windows (eg because of npm's new flat directory) then we can stop using .asar. (Note, I currently dont use .asar in the osx build because it breaks IPFS' daemon-launching. If Windows *has* to use .asar, then I'll need a new solution for IPFS.)
[...]

PR #57, opened 2016-08-15
scan app/node_modules for beaker plugins, not global-modules. Remove …

joshuef:
…global-modules from package.json.

Also removes default inclusion of dat/ipfs if not found in globals.

For https://github.com/pfrazee/beaker/issues/56


pfrazee:
Closing, per the discussion in #56

PR #80, opened 2016-09-05
add open-url support

### LINKED ISSUE #77: support OSX OPEN_URL and protocols
Operation System:
OSX 10.11.6
Beaker Version:
Beaker 0.3

I want to use beaker as my default browser. Currently, it's not possible since it doesn't support "http" and "https" protocol.

`electron-builder` have [built-in support](https://github.com/electron-userland/electron-builder/blob/91addfe4f382059a60fbbd60c2de5aa55d0c42b6/docs/Options.md) for protocol settings. I can try to make a PR for this.

what do you think?

poga:
In order to make this work, `background-process` need to know when `shell-window` is ready, then send `file:new-tab` command to shell-window.

- add 'shell-window-ready' event to indicate shell-window is ready
- add app listener for `open-url` event.
- add `http` and `https` protocol to `package.json` so on OSX we can go `system perferences` -> `general` -> `default browser` and select beaker!

fix #77

### DIFF
--- a/app/background-process/api-manifests/browser.js
+++ b/app/background-process/api-manifests/browser.js
@@ -14,5 +14,9 @@ export default {
   uninstallPlugin: 'promise',
 
   getHomePages: 'promise',
-  getProtocolDescription: 'sync'
+  getProtocolDescription: 'sync',
+
+  getDefaultProtocolSettings: 'sync',
+  setAsDefaultProtocolClient: 'sync',
+  removeAsDefaultProtocolClient: 'sync'
 }
--- a/app/background-process/browser.js
+++ b/app/background-process/browser.js
@@ -56,7 +56,7 @@ export function setup () {
   setTimeout(scheduledAutoUpdate, 15e3) // wait 15s for first run
 
   // wire up RPC
-  rpc.exportAPI('beakerBrowser', manifest, { 
+  rpc.exportAPI('beakerBrowser', manifest, {
     eventsStream,
     getInfo,
     checkForUpdates,
@@ -67,8 +67,27 @@ export function setup () {
     setSetting,
 
     getProtocolDescription,
+    getDefaultProtocolSettings,
+    setAsDefaultProtocolClient,
+    removeAsDefaultProtocolClient,
     getHomePages
   })
+}
+
+export function getDefaultProtocolSettings () {
+  return ['http', 'dat', 'ipfs', 'view-dat'].reduce((res, x) => {
+    res[x] = app.isDefaultProtocolClient(x)
+    console.log(x, res[x])
+    return res
+  }, {})
+}
+
+export function setAsDefaultProtocolClient (protocol) {
+  return app.setAsDefaultProtocolClient(protocol)
+}
+
+export function removeAsDefaultProtocolClient (protocol) {
+  return app.removeAsDefaultProtocolClient(protocol)
 }
 
 export function getInfo () {
--- a/app/background-process/open-url.js
+++ b/app/background-process/open-url.js
@@ -1,26 +1,20 @@
 // handle OSX open-url event
 import { ipcMain } from 'electron'
-var shellReady = false
-var shellWindow
 var queue = []
+var commandReceiver
 
-export function setup (win) {
-  shellWindow = win
+export function setup () {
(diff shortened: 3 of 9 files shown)


poga:
To test this, just checkout the branch, `npm run release`, then drag the packaged app to your application folder. The OS should register the protocol automatically.

Then, beaker will show up in `system preference` -> `general` -> `default browser`.

pfrazee:
Good work. This code is good.

Here's what needs to happen next:

 - Let's move the 'is-shell-window-ready' queue onto the window object itself, so that's it's reusable and has clear ownership by the window.
 - Looking at the code, I think if more than one window is open, they'd all open a URL. We should just choose one window, and then give it focus.
 - Add the new protocol schemes, 'ipfs', 'dat', and 'view-dat'
 - Use http://electron.atom.io/docs/api/app/#appsetasdefaultprotocolclientprotocol-path-args-macos-windows and the other 2 related functions to add a post-install registration UI to the settings page.

Don't feel obligated to make these changes, I'm happy to handle them if you're busy.

poga:
Yeah, I'll update the PR and make these changes tomorrow. Thanks for the comments. I learned a lot.

poga:
I've finished the registration UI. However, I can't find a way to retrieve the result of `setAsDefaultProtocolClient`, the interface is kind of weird now.

`open-url` event can happen before app's `ready` event. Therefore, `window` might not exist when we need the queue. 
I think it should be in its own module. check `app/background-process/open-url.js` for my current implementation. It's simpler now since it doesn't need a `window` argument.

`ipfs, `dat`, and `view-dat` are added to supported protocols.

app/background-process/browser.js
```diff
@@ -67,10 +67,29 @@ export function setup () {
...
+    setAsDefaultProtocolClient,
+    removeAsDefaultProtocolClient,
     getHomePages
   })
 }
 
+export function getDefaultProtocolSettings () {
+  return ['http', 'dat', 'ipfs', 'view-dat'].reduce((res, x) => {
+    res[x] = app.isDefaultProtocolClient(x)
+    console.log(x, res[x])
```
pfrazee: Leftover log()

app/background-process/open-url.js
```diff
@@ -0,0 +1,21 @@
...
+import { ipcMain } from 'electron'
+var queue = []
+var commandReceiver
+
+export function setup () {
+  ipcMain.on('shell-window-ready', function (e) {
+    commandReceiver = e.sender
+    queue.forEach((url) => {
+      commandReceiver.send('command', 'file:new-tab', url)
+    })
```
pfrazee: Need to clear the queue here, otherwise (I think) new windows will open the URLs as well

app/background-process/open-url.js
```diff
@@ -0,0 +1,21 @@
+// handle OSX open-url event
+import { ipcMain } from 'electron'
+var queue = []
+var commandReceiver
+
+export function setup () {
+  ipcMain.on('shell-window-ready', function (e) {
```
pfrazee: This could probably be a .once()

app/background-process/api-manifests/browser.js
```diff
@@ -14,5 +14,9 @@ export default {
...
 
   getHomePages: 'promise',
-  getProtocolDescription: 'sync'
-}
\ No newline at end of file
+  getProtocolDescription: 'sync',
+
+  getDefaultProtocolSettings: 'sync',
+  setAsDefaultProtocolClient: 'sync',
+  removeAsDefaultProtocolClient: 'sync'
```
pfrazee: Lets make these RPC calls promises instead of sync. Even sync calls freeze the thread briefly, and, if they fail, can freeze up the program -- so I try to avoid them.

pfrazee:
Solid work. I really appreciate the quality.

I left a few code-review comments. Again, I'm happy to tackle them myself if you're out of time.

pfrazee:
Ok dont worry about the code-review comments. I'm ready to tackle this, so I'll make the changes myself.

pfrazee:
👍  merged, cheers

poga:
Thanks! 🎉

PR #91, opened 2016-09-11
revert any editing in navbar when ESC is pressed, not just remove focus.

poga:
A minor usability issue here. 

In Chrome, if we press ESC when editing navbar, the value of navbar will be reverted to the page's URL. In safari, we need to press ESC 2 times to achieve the same effect.

This PR will mimic the experience in Chrome, which take 1 ESC key press to revert the changes.

Feel free to close this PR if you don't like this behaviour.


pfrazee:
I ❤️  usability tweak PRs. I'll put it through some tests

pfrazee:
👍  cheers

PR #98, opened 2016-09-21
navbar/autocomplete IME support

### LINKED ISSUE #97: navbar location input act incorrectly when using Chinese IME input.
Operation System:
OSX 10.11.6
Beaker Version:
0.3.2

When typing in Chinese IME, we need to press `<Enter>` to complete a composition. For Example: 

There are two issues in the screenshot.

1. In the screenshot, the composition is not complete.(You can see there's a line under the characters.) Pressing `<Enter>` means I'm confirming the composition is correct, not for submitting the input.
However, `onKeyUp` will be triggered and submit the input, which is incorrect. 

2. The autocomplete suggestion is wrong. My input is `中文`, but autocomplete is based on `ㄓㄨㄥ` which is a partial input of pre-composition Mandarin Phonetic Symbols(Zhuyin).

I can try to put together a PR for this since this is usually hard to notice for non-IME users.

poga:
This PR will fix issues discussed in https://github.com/pfrazee/beaker/issues/97

It moves location submitting into onKeyDown handler to avoid triggered by IME composition event.

fix #97

### DIFF
--- a/app/shell-window/ui/navbar.js
+++ b/app/shell-window/ui/navbar.js
@@ -225,7 +225,6 @@ function render (id, page) {
         class="nav-location-input"
         onfocus=${onFocusLocation}
         onblur=${onBlurLocation}
-        onkeyup=${onKeyupLocation}
         onkeydown=${onKeydownLocation}
         oninput=${onInputLocation}
         value=${addrValue} />
@@ -439,21 +438,6 @@ function onBlurLocation () {
   setTimeout(clearAutocomplete, 150)
 }
 
-function onKeyupLocation (e) {
-  // on enter
-  if (e.keyCode == KEYCODE_ENTER) {
-    e.preventDefault()
-
-    var page = getEventPage(e)
-    if (page) {
-      var selection = getAutocompleteSelection()
-      page.loadURL(selection.url, { isGuessingTheScheme: selection.isGuessingTheScheme })
-      e.target.blur()
-    }
-    return
-  }
-}
-
 function onInputLocation (e) {
   var value = e.target.value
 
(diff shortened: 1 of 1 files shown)


pfrazee:
This is mostly good! But it's showing some incorrect results. There's a weird interaction between the tokenizer and the word-highlighting that results in this:

screen shot 2016-09-21 at 9 38 20 am

Notice how some letters that shouldn't be highlighted, are. For instance, 'beaker/creati' is highlighted, but 'beaker-plugin' is the input.

The results are somewhat weird, however. If I type in 'beaker plugin' (note, no dash) I get this:

screen shot 2016-09-21 at 9 42 27 am

Which means the SQLite portion is not handling the `-` very well. So, there are two avenues to pursue: 1) making sure the word-highlighting in the frontend is correct, and 2) making sure the SQLite search is correct.

poga:
Thanks! I'll keep working on this.

app/shell-window/ui/navbar.js
```diff
@@ -323,17 +324,18 @@ function decorateResultMatches (searchTerms, result) {
     // use the length of the search term
     // (sqlite FTS gives the length of the full matching token, which isnt as helpful)
     let searchTerm = searchTerms[termIndex]
-    let len = searchTerm.length
+    // offset from sqlite is based on bytes. we need to calculate the length of searchTerm in bytes too.
+    let len = byteLength(searchTerm)
```
poga: since the offset returned from SQLite is based on bytes. this function will decorate wrong segments since it's splitting segments based on character length. Here is an example(highlighted with red so it's easier to see):



This commit fixed the problem by splitting segments based on bytes.

poga:
Actually, I think it's better to keep this PR clean and only fix the IME problem. So I removed all search-related commit from it.

I'll open another issue to track search-related problems.

poga:
NVM I messed up this one. will open another one if I can isolate a solution

pfrazee:
Ok sounds good

PR #109, opened 2016-09-27
Chinese IME support on navbar

### LINKED ISSUE #97: navbar location input act incorrectly when using Chinese IME input.

poga:
This is a simplified PR for https://github.com/pfrazee/beaker/issues/97#issue-178170054

The PR only fix issues related to Chinese IME on `navbar`.

Issues mentioned at https://github.com/pfrazee/beaker/pull/98#issuecomment-248633956 can be reproduced on current master. I'll open another issue to track those problems.

fix #97

### DIFF
--- a/app/shell-window/ui/navbar.js
+++ b/app/shell-window/ui/navbar.js
@@ -225,7 +225,6 @@ function render (id, page) {
         class="nav-location-input"
         onfocus=${onFocusLocation}
         onblur=${onBlurLocation}
-        onkeyup=${onKeyupLocation}
         onkeydown=${onKeydownLocation}
         oninput=${onInputLocation}
         value=${addrValue} />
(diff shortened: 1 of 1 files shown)


pfrazee:
Good deal, thanks @poga

PR #124, opened 2016-10-09
"add options for bulk deleting history"

taravancil:
This adds the ability to bulk delete history items from today, this week, this month, or from all time.

### DIFF
--- a/app/background-process/api-manifests/history.js
+++ b/app/background-process/api-manifests/history.js
@@ -4,5 +4,6 @@ export default {
   getMostVisited: 'promise',
   search: 'promise',
   removeVisit: 'promise',
-  removeAllVisits: 'promise'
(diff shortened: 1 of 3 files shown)


pfrazee:
Looking great! I'll need to take a look at visit_fts in a second, and see how I think we should handle the new deletion model.

After https://github.com/beakerbrowser/beaker/pull/124/files#diff-222fd164f533a76625e85b25af6a3dbaR143 do we need to fetch entries, to make sure the screen isn't (incorrectly) empty? Eg, if the history screen was showing exactly all of one day's history, the user clears last 24 hrs, and then it'd be empty: when really it should show the entries from >24hrs.

pfrazee:
👍  Merged. Nice work!

PR #153, opened 2016-10-25
Smarter History

### LINKED ISSUE #70: Smarter history logging
Some improvements we can make:

 - Dont log multiple consecutive visits to the same site. Maybe: if the URL was visited in the last hour, just update its timestamp.
 - If the title changes within 15 seconds of the load, but the URL does *not* change, then use the changed title. Some pages will display something like "Loading..." in the title, then change to the content's title when ready.

poga:
* don't log multiple consecutive visits to the same url
* update title in history if it's changed within 15s after page load and the URL doesn't change
* sort history by ts instead of rowid

fix #70


pfrazee:
Really happy you did this! Thanks so much.
👍  Code is good. Merging

PR #187, opened 2016-11-16
change tab traverse hotkey to be consistent with other major browsers

poga:
All major browsers (Chrome/Safari/Firefox) use `cmd+{` and `cmd+}` as the hotkey to go to next/previous tab.

We should use the same hotkey for accessibility.


pfrazee:
Oh, you're right. And it looks like `cmd+[` and `cmd+]` do back/forward in Chrome and Firefox. I should add that.

Good catch, thanks for the fix.

PR #191, opened 2016-11-20
Dat API 2.5

pfrazee:
A few rewrites and simplifications of how dats are managed internally.

 - [x] **Internal Data** Replace "claims" with isSaved and isHosting flags
 - [x] **Hosting** Remove serve() and unserve() methods. Set `isHosting` to true by default, for created archives. Only allow the user to control hosting via Beaker's builtin UIs.
 - [x] **Write Perms** Replace write-permission rules with a list of authorized writers. Include the creating app by default. For unauthorized apps, prompt the user for every write ("This site would like to write /bar.png (5 mb) to 'My Pics.'"). Include an 'Always Allow' checkbox, which adds the writing site to the target dat's list of authorized writers.
 - [x] **Quotas** Change file-usage limits to be per-Dat, instead of per-App.
 - [x] **Deletion** Change `deleteArchive()` to prompt the user before execution.


pfrazee:
👍  ready to go

PR #196, opened 2016-11-25
default localhost to http

### LINKED ISSUE #195: Default to http:// on localhost
It's currently defaulting to https, which will almost never be the right choice

poga:
fix #195


pfrazee:
👍  thanks!

PR #197, opened 2016-11-26
implement archive history view

### LINKED ISSUE #172: Archive history view to beaker:archive
Same content as in https://github.com/mafintosh/dat-ls

The beaker:archive UI should let the user see the change history, perhaps via a selectable tab.

poga:
Add an archive history tab to `beaker:archive` based on https://github.com/mafintosh/dat-ls

close #172


pfrazee:
👍  💯  🎉  Nice work! You nailed it

PR #234, opened 2016-12-30
select navbar content even when it's already focused

zeke:
This PR makes the <kbd>CmdOrCtrl</kbd><kbd>l</kbd> shortcut select the contents of the address/location/nav/wonder/omni bar, even if it's already focused. This behavior is consistent with Chrome, Firefox, and Safari.

Please let me know if this should have a test somewhere!

beaker-select


pfrazee:
Awesome, thanks!

Issue #229, opened 2016-12-27
Render hashes in the URL bar with pretty-hash


pfrazee:
The hash is not particularly useful to see. It's more useful to see the path. So, show a truncated version. Then, when the user focuses into the URL bar, re-expand to the full hash.

zeke:
https://github.com/pfrazee/pretty-hash 💅

PR #242, opened 2017-01-01
Render hashes in the URL bar with pretty-hash, and gray the path

pfrazee:
Closes https://github.com/beakerbrowser/beaker/issues/229

screen shot 2017-01-01 at 4 05 38 pm

@zeke I decided to keep the shortened hash as-is, for now, so that it's very clear that a truncation is happening


zeke:
:+1:

PR #243, opened 2017-01-02
Toolbar fixes

pfrazee:
1. Fixes a problem with the add button rendering in the toolbar
2. Nicer font on the location input:

screen shot 2017-01-01 at 7 25 28 pm


pfrazee:
I ended up reverting the font, didnt like it enough

PR #247, opened 2017-01-04
Add a Host Dat Button to Navbar

rjsteinert:
Just getting this started. Need to finish the click handler.


pfrazee:
This is closed by https://github.com/beakerbrowser/beaker/pull/262

Any site which is saved for offline will be served on the network.

Issue #227, opened 2016-12-21
Support dat shortnames via TLS?


cjb:
The current DNS shortnames have no authentication, so they aren't usable in production.  To be comfortable using shortnames, we need a decentralized method of assigning dat hashes to shortnames that has authentication.

I'd guess that TLS is the best way of doing this, even though it's not ideal, since it's already out there, decentralized and authenticated.  The path looked up could be:

https://hostless.website/.well-known/dat-url

A TLS result would have to take precedence over a DNS TXT result, otherwise we haven't improved the security of shortnames.

Thanks!

pfrazee:
This may be a smart decision. It'll be a while before we can add authentication to the dat protocol.

maxogden:
been a while since ive looked at the spec but would this technically be WebFinger?

pfrazee:
@maxogden it is close. Webfinger uses the .well-known path-space, but that's shared by a number of specs.

pfrazee:
Another option is to `HEAD /` and include a header in the response

maxogden:
@pfrazee ah ok. i didnt like WebFinger a few years ago cause it was XML only but was hoping maybe something better had come out

also, if we're cool inventing specs (vs using dnssec or whatever) we could do a simple stateless message based encrypted udp thing (similar to dns but using sodium)

pfrazee:
@maxogden You should give Webfinger another look, it's all json now.

We can invent as much as we like within TLS. We need the cert-based authentication.

cjb:
> Another option is to HEAD / and include a header in the response

I think we'd want something that most website owners can achieve.  "Put a file here" works approximately anywhere (think e.g. GitHub Pages, Heroku), but anything involving changing HTTP configuration doesn't.

pfrazee:
Good point

cjb:
Oh, just to get back to something from earlier:

> It'll be a while before we can add authentication to the dat protocol.

How would that help with forward resolution of shortnames?  I'm having trouble imagining that.  (Feel free to point me towards old design docs or whatever.)

pfrazee:
We havent explored it yet in depth. Specifically, we'd need to authenticate a DNS mapping.

One way to do it is DNSSEC, except the crypto community doesn't love it (weak RSA keys for one, centralization of authority, other easons) and it doesn't have extensive support by TLDs. Another way would be to get certs issued by CAs which can be placed in Dats, but that's obviously going to require some lobbying.

[...]

PR #275, opened 2017-02-03
Implement Dat DNS-over-HTTPS

pfrazee:
Tasks

 - [x] Resolve DNS over HTTPS.
 - [x] Implement an in-memory cache.
 - [x] Discover Dat alternatives to HTTPS sites and prompt the user to redirect.

Spec: https://github.com/beakerbrowser/beaker/wiki/Authenticated-Dat-URLs-and-HTTPS-to-Dat-Discovery

screen shot 2017-02-02 at 6 47 11 pm


pfrazee:
The "prompt to redirect" is a very simple P2P icon in the right side of the navbar. It will need work in the future.

maxogden:
👍  this is awesome

pmario:
I like this possibility, because of the human readable format. ... BUT what about "link-rot", which is a major problem. 

Let's say I have a link eg: `dat://beakerbrowser.com/blog/` but the `beakerbrowser.com` domain doesn't exist anymore. So `https://beakerbrowser.com/.well-known/dat` doesn't return the dat-key anymore. 

 - Where and how do I get a valid dat-key?
 - Even worse. What if the domain name belongs to someone else now?

just some questions that have come up. ... IMO none of them have been asked in #227

pfrazee:
I think link rot is an inevitable tradeoff of an aliasing system. Unless the aliases are globally unique and permanent, you get link rot.

I personally wouldnt want an immutable aliasing system. Just yesterday some glitch caused me to lose the keypair that I was publishing dat://beakerbrowser.com with, so I had to update the DNS entry.

pmario:
Yea. ... After posting it, I thought a little bit about it. eg: If one uses "prettylinks" in publications eg: PDF, the best practice should be to also publish the dat-link as a footnote, or somewhere else, where a user can find it. ... Not perfect but imo a usable workaround.

pfrazee:
@pmario I think that's a pretty reasonable tradeoff. At least now it's *possible*, which is a big step up

pfrazee:
One thing I'd like to do is enable site "snapshots", which uses Dat's existing ability to create static content-addressed archives (as opposed to the typical "live pubkey-addressed" archives). The idea would be, you nav to a site and say "New snapshot," and that'd be similar to forking: you create a new archive with the current content, but it's static and content-addressed.

karissa:
The only thing I'm thinking is a problem here is that the swarm will become less healthy if the hash isn't able to be looked up by the live key. Ideally, the snapshot would be a subset of the live key so that any clients broadcasting the live key or the snapshot can share that with eachother.

pfrazee:
@karissa that's a good point. If we realize that needs to be a priority, we can use versioned URLs instead.

pfrazee:
(And the plus side to versioned URLs is, we support them already!)

PR #304, opened 2017-02-20
focus the webview after the dom-ready event

### LINKED ISSUE #261: <input> element focus() has different behaviour as with chrome browser
Operation System: ubuntu 16.10
Beaker Version: master - commit 77e9f72

To reproduce: 

Try this MTC with chrome and then with beaker
 - https://jsfiddle.net/7t04wszc/15/   with autofocus attribute
 - https://jsfiddle.net/7t04wszc/16/   without autofocus.

related to: https://groups.google.com/d/msg/beaker-browser/9m8pQXshSeo/bEAs3OH2AwAJ but my MTC is a bit different, since the TW UI is much more complex.

may be related to: #253

taravancil:
This focuses the webview after the dom-ready event fires, which fixes
a bug where inputs with the autofocus attribute would not focus unless
the webview was already focused.

Fixes #261

### DIFF
--- a/app/shell-window/pages.js
+++ b/app/shell-window/pages.js
@@ -451,6 +451,7 @@ function onDomReady (e) {
   var page = getByWebview(e.target)
   if (page) {
     page.isWebviewReady = true
+    page.webviewEl.focus()
     zoom.setZoomFromSitedata(page)
   }
 }


pfrazee:
👍

PR #318, opened 2017-03-02
Improve the browser UI

pfrazee:
- Add vibrancy (semi-transparency) (OSX)
- New icons
- Modified tab coloring


ThatOneGamer999:
Ran successful build on Ubuntu 16.04 LTS, Did not see any transparency (of course, may be OSX only).  
No glitches with the toolbar, and works perfectly.  
image

pfrazee:
@ThatOneGamer999 thanks for reporthing. You're right, the actual feature we're using here is the 'vibrancy', and that's OSX specific. We'll have to find a way to make Beaker look ✨  on non-OSX platforms.

PR #330, opened 2017-03-05
V0.7 better chrome

taravancil:


pfrazee:
💯  I rebased into v0.7, clsoing

PR #333, opened 2017-03-05
V0.7 downloads

taravancil:


pfrazee:
👍 👍 👍

ideocrat:
Windows 7 Home Premium Service Pack 1
Beaker Browser 0.7.11

You should be able to change the location of downloads in the settings or downloads tab.
One minor thing when opening the downloads tab in windows it shouldn't say "Show in Finder" since finder is for Mac. It should probably say "Show in folder" or "Show in explorer" which is better understood by Windows users.
capture

taravancil:
@ideocrat your screenshot is out of date, we have a new release out

PR #336, opened 2017-03-06
Show current directory path in site library

pfrazee:
screen shot 2017-03-06 at 2 08 36 pm
screen shot 2017-03-06 at 2 08 48 pm


pfrazee:
@taravancil this 👌  with you?

PR #358, opened 2017-03-20
V0.7 touchups

taravancil:


app/background-process/protocols/beaker.js
```diff
@@ -82,6 +82,9 @@ function beakerServer (req, res) {
     return cb(403, 'Forbidden')
   }
 
+  if (requestUrl === 'beaker://assets/error-page.css') {
```
pfrazee: We need to double-check the content-security policies for this. I think, for instance, the dat:// sites arent allowed to serve styles from the beaker: protocol right now.

pfrazee:
Awesome work

PR #391, opened 2017-04-08
Builtin toasts

taravancil:
This adds a reusable toast component that can be used on builtin pages. Toasts fade out and are removed from the DOM after just over a second.

```
toast.create('A toast message.')
```
screen shot 2017-04-08 at 4 04 27 pm


pfrazee:
👍 👍 👍

PR #392, opened 2017-04-09
Enhanced error page

taravancil:
This changes the error page to handle:

* failed DNS resolution,
* failed TLS connection, and
* other errors with a general fallback page


pfrazee:
👍 👍 👍

PR #393, opened 2017-04-11
open beaker://library instead of beaker://network

frabrunelle:
I noticed that clicking on the number of peers on a `dat://` site currently opens `beaker://network`, which was recently renamed to `beaker://library`.

screen shot 2017-04-11 at 12 15 29 am

I changed the `onClickPeers` function to open `beaker://library` instead 🙂

I'm super excited for Beaker 0.7 😃 Cheers!


pfrazee:
Thanks!!

PR #408, opened 2017-04-19
Dont link to library from URL bar peer count

taravancil:


pfrazee:
👍 👍 👍

joehand:
Ah yes, I was confused by this =).

PR #411, opened 2017-04-27
Redesign of Dat storage, better OS integration, and removed editor

pfrazee:
This is a pretty big set of changes that I'll write a post about, at some point. Here's a short background for now: https://toot.cafe/@paul/310219

This PR:

 - Adds the concept of a "Staging" area to Dats
 - Adds APIs for diffing, committing, and reverting changes to Dats that are in the staging area
 - Now prompts the user to select a staging folder when Dats are created or saved
 - Removes the builtin editor


pmario:
> - Removes the builtin editor

interesting

taravancil:
@pmario https://toot.cafe/web/statuses/321416

PR #430, opened 2017-05-12
Update security UI on error pages

taravancil:
This fixes the security UI on error pages and adds a "Not secure" label on pages with HTTPS errors
screen shot 2017-05-12 at 11 26 11 am
screen shot 2017-05-12 at 11 26 19 am

### DIFF
--- a/app/shell-window/ui/navbar.js
+++ b/app/shell-window/ui/navbar.js
@@ -127,6 +127,7 @@ function render (id, page) {
   const isViewingDat = page && page.getURL().startsWith('dat:')
   const siteHasDatAlternative = page && page.siteHasDatAlternative
   const gotInsecureResponse = page && page.siteLoadError && page.siteLoadError.isInsecureResponse
+  const siteLoadError = page && page.siteLoadError
 
   // back/forward should be disabled if its not possible go back/forward
   var backDisabled = (page && page.canGoBack()) ? '' : 'disabled'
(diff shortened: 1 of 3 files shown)


pfrazee:
Are you sure that every failed response should be treated as a security error?

taravancil:
Hm, I can't think of any situation where we'd ever want to label a failed response as "Secure". I don't feel like this UI indicates there's a security error, do you?

pfrazee:
> I don't feel like this UI indicates there's a security error, do you?

It says "Not secure"

>  Hm, I can't think of any situation where we'd ever want to label a failed response as "Secure"

A 404 is a secure response

taravancil:
The "Not secure" label is only on sites that *also* got an insecure response

taravancil:
The only 404s we handle is on builtin pages right? so `beaker://notfound`? In that case we don't show any security UI, it's treated like any other `beaker://` URL

pfrazee:
> The "Not secure" label is only on sites that also got an insecure response

Oh I see.

Still -- why change the protocol icon for a 404?

taravancil:
That's the icon both FF and Chrome use for 404s, DNS errors, etc.. If we feel like we need to use a different icon than the HTTPS one, we could use the `.exclamation-triangle` icon for HTTPS errors

pfrazee:
That's not quite accurate. If you go to https://foo.hashbase.io/ you get a 404 and chrome renders the normal secure protocol icon

taravancil:
You're right. We do the same thing too! 
screen shot 2017-05-12 at 12 00 32 pm

taravancil:
A 404 is handled by the site that 404s right? That's not considered a failed load on our end.

pfrazee:
Does this PR do that?

taravancil:
Yeah, that screenshot is on this PR's branch

pfrazee:
Well... hmm ok then! I thought that would populate `page.siteLoadError`. Merging

taravancil:
👍

PR #436, opened 2017-05-12
Remove bkr support

pfrazee:
**Why was bkr originally created**

Prior to 0.7, the Dat sites had no direct file-system integration. The data was stored in Beaker's internal storage, and the only way to access it was within the app.

We needed a way to pull the files into the FS so that an external editor could be used.

**Why bkr is being removed now**

In 0.7, Dat sites are now saved in a folder that the user controls. That makes bkr redundant. Since we're a small team, we want to keep the number of tools we maintain small too. So, we're going to remove bkr support now.


millette:
I hadn't had a chance to use bkr before it was removed. But I've been toying with Beaker and my prefered editor and found it just works. Not sure what I would be missing from bkr, so there.

PR #439, opened 2017-05-13
Update how toggleables can be closed

taravancil:
This makes it possible to close a toggleable element with ESC and clicking anywhere on the page.

@pfrazee, do we also need to remove the `keyup` event listeners like I do for the `document`'s `click` event?

### DIFF
--- a/app/builtin-pages/com/toggleable.js
+++ b/app/builtin-pages/com/toggleable.js
@@ -1,4 +1,5 @@
 import * as yo from 'yo-yo'
+import { findParent } from '../../lib/fg/event-handlers'
 
 // globals
 // =
@@ -13,6 +14,7 @@ var toggleState = {}
 // give class .toggleable, .toggleon, or .toggleoff to trigger
 // include data-toggle-on="event", where `event` sets what triggers toggle (default click)
 // include data-toggle-id if you want to keep the toggle state across renderings
+
 export default function toggleable (el) {
   var id = el.dataset.toggleId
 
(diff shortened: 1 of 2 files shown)


pfrazee:
Glad you hit this!

> do we also need to remove the keyup event listeners like I do for the document's click event?

No, actually what you're doing *should* be a logic error. This toggleable helper doesn't trigger yoyo rendering; it just toggles an 'open' class to show/hide rendered elements. That means the click handler that you're removing has no guarantee to be set again when the user reopens the toggleable, and so (if I'm reading the code right) you should only be able to close the toggleable by clicking outside of it once.

There's a better way to handle both the keyup and click-on-document events. Outside of any function, attach event listeners to the document for both keyup and click.

 - For the keyup handler, just check the keycode, and if it's escape, call `closeAllToggleables()`
 - For the click handler, use [findParent](https://github.com/beakerbrowser/beaker/blob/f4a4e224a8d8b612428c7de0dccbab882d3e01ea/app/lib/fg/event-handlers.js#L18) to see if the `e.target` is inside of a `.toggleable-container`.  If it **isnt**, then call `closeAllToggleables()`

If that's unclear, ask me for help.

taravancil:
Ah I see!

Follow up question: since we're listening for a document click, then we don't need a click listener on any other elements like we do [here](https://github.com/beakerbrowser/beaker/pull/439/files#diff-1573fd010971036865d2b9c0d6fa3a47L26), right?

Unless of course `el.dataset.toggleOn` is specified, then we should.

app/builtin-pages/com/toggleable.js
```diff
@@ -31,13 +33,16 @@ export default function toggleable (el) {
...
+
   function onToggle (e) {
     e.preventDefault()
     e.stopPropagation()
     var newState = !el.classList.contains('open')
     closeAllToggleables()
     if (newState) {
       el.classList.add('open')
+    } else {
+      document.removeEventListener('click', onToggle)
```
pfrazee: This isn't needed. We never add onToggle to document's click

app/builtin-pages/com/toggleable.js
```diff
@@ -57,6 +62,7 @@ export default function toggleable (el) {
     e.preventDefault()
     e.stopPropagation()
     el.classList.remove('open')
+    document.removeEventListener('click', onToggleOff)
```
pfrazee: Same deal, isn't needed

taravancil: whoops! sorry i missed that

pfrazee:
👍

PR #446, opened 2017-05-14
Make the library responsive

taravancil:
I know we said we weren't going to prioritize this, but it was becoming enough of a pain point for me that I decided to do it anyway 😄 
screen shot 2017-05-14 at 2 48 41 pm
screen shot 2017-05-14 at 2 48 49 pm


pfrazee:
:+1:

PR #447, opened 2017-05-16
v0.7 onboarding

taravancil:


app/background-process/protocols/beaker.js
```diff
@@ -177,20 +176,74 @@ async function beakerServer (req, res) {
...
-    // TODO re-enable
-    /*let status = await getUserSetupStatus()
-    if (status !== 'finished') {
-      // serve the setup if the user isnt finished with setup
-      return cb(200, 'OK', 'text/html', path.join(__dirname, 'builtin-pages/setup.html'))
-    }*/
     return cb(200, 'OK', 'text/html', path.join(__dirname, 'builtin-pages/start.html'))
   }
+  if (requestUrl === 'beaker://tour/') {
+    return cb(200, 'OK', 'text/html', path.join(__dirname, 'builtin-pages/setup.html'))
```
pfrazee: Can we rename setup.html to tour.html?

app/builtin-pages/views/start.js
```diff
@@ -37,6 +38,8 @@ async function setup () {
     userProfile.title = 'Your profile'
   }
   settings = await beakerBrowser.getSettings()
+  userSetupStatus = await beakerBrowser.getUserSetupStatus()
```
pfrazee: Let's not use userSetupStatus for this, because we'll need that later for setting up a profile. Instead, use the technique that `latestVersion` uses in start.js (the siteData DB). Create a new entry, like 'watched-tour'.

app/background-process/protocols/beaker.js
pfrazee: Same story with setup.js

taravancil: yep!

pfrazee:
👍  great work

PR #449, opened 2017-05-16
Changes to sites folders and dat.json handling

pfrazee:
After some QA, we've hit on some improvements that this PR should make:

- [x] Save all sites to a `~/Sites/{title}` folder by default, rather than always prompting the user where to save
- [x] Show the site's folder more prominently in the library, and give a more prominent "Open folder" button/link
- [x] Include a tool in the library to change the folder after the initial creation/save
- [x] Don't allow special OS folders to be used as the location of sites (eg ~/Documents, ~/Music, etc)
- [x] Allow the sites folder to be overridden by an environment variable
- [x] Don't read/write the dat.json in the staging; instead, just read/write it directly to the archive
- [x] Update tests
- [x] Restore old create/fork tests, now that a file-picker modal isnt required (yay!)


taravancil:
- [ ] Automatically publish changes when forking
- [ ] Change folder should open sites folder
- [ ] Empty state for file listing

PR #493, opened 2017-06-01
Add support for HTTP sites in beaker://view-source

taravancil:


app/background-process/browser.js
```diff
@@ -109,6 +111,19 @@ export function setup () {
   })
 }
 
+export function fetchBody (url) {
+  return new Promise((resolve) => {
+
+    var http = url.startsWith('https') ? require('https') : require('http')
+
+    http.get(url, (res) => {
+      var body = ''
```
pfrazee: Add `res.setEncoding('utf8')` here to be sure it comes out a string

PR #511, opened 2017-06-06
Add {stat:} opt to DatArchive.readdir()

pfrazee:
Adds an option to `DatArchive` API's `readdir()`

```js
await archive.readdir('/', {stat: true})
/* => [
  {name: 'foo', stat: {...}},
  ...
]*/


pfrazee:
Is also used in the library to prove performance of the site exploration.

Issue #508, opened 2017-06-06
"Download full Dat" control


pfrazee:
Currently there's no way to request that Dat files are downloaded, except by clicking on them. It would be really great to have a tool that lets you trigger downloads of some/all of the archive.

todrobbins:
I like the simplicity of Dropbox's UI on macOS:

![screen shot 2017-06-08 at 8 46 27 am](https://user-images.githubusercontent.com/158590/26934832-b07a8f0c-4c27-11e7-80c1-dc9b71c8dc01.png)

Right click and select via a contextual menu seems straight forward enough. Thoughts?

pfrazee:
Oh hey, excellent suggestion. We'll keep that in mind.

pfrazee:
Done with the recent sidebar changes

PR #512, opened 2017-06-07
Improvements to dat downloading and historic-cache management

pfrazee:
Currently, Beaker handles syncing a 'saved' Dat by using the staging area's diff and revert(). We discovered the diff handles large dats very poorly. The solution also caused all dat files to be duplicated, which was subpar. In this PR, we add a faster algorithm for syncing latest, and also improve the deletion of historic files in a Dat.

Still TODO

 - [ ] Unlink from latest dir
 - [ ] A way to change the storage configuration when a Dat's localPath changes
 - [ ] Update the garbage collector to remove historic files
 - [ ] Automatically download latest, perhaps by using hyperdrive `latest=true` or by copying key parts of that algorithm
 - [ ] Turn our storage code into a module
 - [ ] Update multi-random-access to tell whether a read or write is being done, and then only replace hard links if a write is occurring


app/background-process/networks/dat/storage.js
```diff
@@ -0,0 +1,138 @@
...
+        if (err) return cb(err)
+        if (!node) return cb(new Error('Could not locate data'))
+
+        // provide the storage
+        var historicFilePath = path.join(metadataDir, 'content', 'files', `${node.name}.${st.offset}`)
+        cb(null, {
+          start: st.byteOffset,
+          end: st.byteOffset + st.size,
+          storage: raf(historicFilePath)
+        })
```
mafintosh: my gut tells me we should wait on calling the callback until after the hardlinker has run ... for safety reasons

pfrazee: You think? No code should rely on those hardlinks: they exist purely for the end-user to be able to access the files in a convenient place

mafintosh: I think this might bite us in the future when we start doing deletes otherwise

pfrazee:
We gave this some thought and decided the core UX was off. The sites directory is a bit confusing.

We're instead going to aim for an experience similar to what's suggested in https://github.com/beakerbrowser/beaker/issues/508#issuecomment-307128116, and then add controls to download sites as a zip.

PR #527, opened 2017-06-13
Dat sidebar (WIP)

taravancil:
This adds the Dat sidebar on Dat websites.

There are still some TODO items, but I'll leave them to you @pfrazee since they all are related to managing the open/closed state of the sidebar

- [x] Update the buttons
- [x] Fix .files-list rendering issue

### DIFF
--- a/app/background-process/protocols/beaker.js
+++ b/app/background-process/protocols/beaker.js
@@ -205,6 +205,15 @@ async function beakerServer (req, res) {
   if (requestUrl === 'beaker://settings/main.js') {
     return cb(200, 'OK', 'application/javascript', path.join(__dirname, 'builtin-pages/build/settings.build.js'))
   }
+  if (requestUrl === 'beaker://dat-sidebar/main.js') {
+    return cb(200, 'OK', 'application/javascript', path.join(__dirname, 'builtin-pages/build/dat-sidebar.build.js'))
+  }
+  if (requestUrl === 'beaker://dat-sidebar/main.css') {
+    return cb(200, 'OK', 'text/css', path.join(__dirname, 'stylesheets/builtin-pages/dat-sidebar.css'))
+  }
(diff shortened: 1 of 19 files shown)


pfrazee:
Looking great. A couple UI/UX notes:

 - The font is just too small. Some of it feels right (the buttons) but I think the content text needs to be bigger, and we should widen the sidebar to make that work.
 - We should bold the title
 - The top buttons should be fixed and not scroll, same as the bottom button

For future PRs:

 - We need to rethink the save/remove button.
 - For owners, we'll need btns to add files, open the folder, and copy the folder path
 - For non-owners, we'll need a btn to download a zip of the the archive

pfrazee:
Just reviewed the code, that's all good work

pfrazee:
Progress report

sidebar

taravancil:
This is awesome! Though I don't think we can use that icon. It looks like a play button.

taravancil:
Is there anything else you need to work on or can I take over from here?

pfrazee:
No I've still got some TODOs

taravancil:
Progress report:

This has come along soo well!

screen shot 2017-06-15 at 7 14 29 pm 2



There are few minor TODOs that I think will seal the deal. I'm OK with merging before we do these, but I wanted to write them down so we don't forget.

- [ ] Close the sidebar automatically when the user navigates to away from the given archive by clicking a link from *within* the sidebar. This includes the network debugger and library view.
- [ ] Alternatively, we could consider closing the sidebar when the user navigates to any non-Dat website. We would still allow the user to open the sidebar on a non-Dat site.
- [ ] Gosh the rendering is slow :cry: I wonder if it's worth trying to mock the content to make it seem faster?

pfrazee:
Ok, I'm going to go ahead and merge. Nice work!

PR #588, opened 2017-07-09
UI improvements to the Dat Archive Picker

pfrazee:
Updates the UI of the modal to the `DatArchive.selectModal()` API

screen shot 2017-07-09 at 2 33 53 pm

---

Previous version:


taravancil:
This turned out really nice! So much simpler than before.

PR #589, opened 2017-07-10
Web Terminal (WIP)

pfrazee:
This is a work in progress. It's currently delayed while we focus elsewhere, but if somebody wants to pick up the PR, we'd be 100% behind that.

Spec: https://github.com/beakerbrowser/beaker/wiki/WebTerm

**TODOs**

 - [ ] Create the program execution sandbox
 - [ ] Create the program RPC system
 - [ ] Add the user root archive
 - [ ] Load the terminal environment from files in the user root
 - [ ] Check that the destination exists before doing a CD
 - [ ] More builtin commands
 - [ ] History scrolling
 - [ ] Terminal autocomplete


webdesserts:
@pfrazee I tried to run this branch today but I wasn't able to use the `term://` protocol. It would ask me for some sort of permission and then go nowhere. Is that how I'm supposed to access the terminal or was there another way?

pfrazee:
@webdesserts Are you sure the term protocol was successfully registered? If it asks permission it's because the protocol didnt register

webdesserts:
@pfrazee I don't know what that means 😅

Are you asking if it was successfully registered as a new protocol in the code? I'm guessing you're referring to [this](https://github.com/beakerbrowser/beaker/pull/589/files#diff-69c10f6428367a58b31880dc99ac9b67R25)? If something's broken there I'll try to debug it after work. Just didn't know if I was accessing the terminal in the ui correctly or not.

pfrazee:
> I'm guessing you're referring to this?

Yep! See if there are any errors there.

webdesserts:
@pfrazee btw, I figured out what the issue was. Apparently I did something like

`git checkout -b upstream/clj`
rather than
`git checkout -b cli upstream/cli`

working with multiple remotes is a pain 😅

pfrazee:
Hah ok cool

pfrazee:
I've factored out the code into a userland site, https://github.com/pfrazee/webterm

PR #594, opened 2017-07-11
Fix up styles on the write permissions prompt

pfrazee:
Before:

screen shot 2017-07-11 at 11 52 36 am

After:

screen shot 2017-07-11 at 11 51 48 am

The link was always there ("My Web site" should be blue in the "before" pic) so that's what I set out to fix. Then I figured I'd touch up the button styles while I was at it.


taravancil:
Looks great!

PR #596, opened 2017-07-11
Improvements to dat configuration

pfrazee:
We felt we could make the process of saving sites more clear. Instead of "Sync for offline," we now have a button to "Add to library," and then additional settings for configuring the auto-downloader.

updated-sidebar

We also have a UI planned for configuring auto-uploading, but this is blocked until support for no-upload is added to hypercore's replicate() method:

uploader_todo


pfrazee:
This PR also adds the "Delete downloaded files" tool to the sidebar's setting tab

pfrazee:
Updated the library to match:
screen shot 2017-07-11 at 6 58 56 pm
screen shot 2017-07-11 at 6 59 15 pm

Issue #591, opened 2017-07-11
Option to populate forkOf in dat.json


millette:
Currently, forks will leak the IDs of the parent dats. One might want to share a forked dat without revealing its origins.

Could be an option in the fork dialog and api method.

Question is, should it default to current behavior or is that a bad leak in a dat context?

See also #566 on naming (fork, clone).

pfrazee:
We have a similar question about `createdBy` in https://github.com/beakerbrowser/beaker/issues/544, and I'm wondering in both cases if they're ever a good idea. We might need to just remove both.

millette:
```forkOf``` and ```createdBy``` could probably be removed in favour of greater user control over the dat.json file.

millette:
Oh, I missed #544. These issues could be merged, #591 (this) could be closed as a dupe.

Issue #544, opened 2017-06-16
rethink createdBy in dat.json


taravancil:
Beaker adds a `createdBy` property to `dat.json` which indicates from where the archive was created. If it's another archive, its value is the key to that archive.

This prohibits an application from creating a new archive in order to share only some of its files secretly.

pmario:
Is this a security concern, or a privacy concern?

pfrazee:
Privacy. An example is, you have a site with lots of private files, and you add a script to create new archives to share a subset of the files. Whoops, just gave away the private site.

pmario:
>Privacy. 

OK

I can see the reasoning. I was just confused by the label: "security feature" ... I'm not sure if it's worth a new label. eg: "privacy feature"

PR #598, opened 2017-07-12
Deprecation: Remove createdBy and forkOf from dat.json manifests

pfrazee:
Closes https://github.com/beakerbrowser/beaker/issues/591 and https://github.com/beakerbrowser/beaker/issues/544

We decided this metadata offers pretty little value, but it does create the potential for accidentally leaking private information.


pfrazee:
cc @jimpick I noticed you used this metadata in one of your apps. LMK if this deprecation is a big problem!

rjsteinert:
Perhaps this makes sense as opt-in.

pfrazee:
@rjsteinert yeah perhaps we'll bring it back in that form later

PR #613, opened 2017-07-20
Dat page menu

pfrazee:
Adds a menu to the navbar of dat pages, to do common actions. This is to replace the dat sidebar with something a bit faster, more convenient, and less dauting
screen shot 2017-07-19 at 11 01 56 pm


taravancil:
I love this! I'm really glad these contextual items are somewhere accessible again.

Issue #552, opened 2017-06-20
How do we feel about the network-access policies for dat:// apps?


pfrazee:
Currently, dat:// sites have to ask permission to access a remote host. You can't embed assets unless they're served over dat://, and you can't Ajax to a remote host.

To relax that, you have to use the [permissions API](https://beakerbrowser.com/docs/apis/permissions.html) to request special access.

```js
var res = await navigator.permissions.request({
  name: 'network',
  hostname: 'github.com'
})
res.status // => 'granted'
```

We're wondering if that's too much of a pain for developers? The CSP limitations can be a nasty surprise.

## Arguments for "no network access by default"

Disabling embeds and connections from HTTPS makes it much harder to track users as they browse the Web. Something look Google Analytics simply doesn't work.

Further, the sandbox makes it possible to trust Dat sites that handle personal data, because it's hard for them to send the data out (exfiltrate) via Ajax.

## Arguments against "no network access by default"

The P2P swarm is not exactly high-privacy right now, so perhaps stopping the trackers is more trouble than it's worth!

Also, writing a Dat app is irksome when I can't leverage resources from CDNs, or read data from HTTPS sources.

## Options

I'll comment with some options, that we can then vote on

pfrazee:
Option 1: Remove this policy so our lives can be easier as devs.

pfrazee:
Option 2: Make this policy user-configurable. A paranoid like myself would then turn it on.

pfrazee:
Option 3: Add a manifest.json entry for requesting permission. That way, it's simpler to write apps that access remote resources.

pfrazee:
We can choose more than one of these. What's the community think? cc @taravancil @mafintosh @joehand @pmario @frabrunelle @jondashkyle

joehand:
My biggest argument for "no network access by default" is that dat stuff will *always* work offline. It is a constraint that changes how things are built.

pfrazee:
Good point

pmario:
I think it's the users decision and not the developers. So a manifest seems to be OK, but the user has to have the final say. Especially with the `*` wildcard being a possibility. ... Which I wouldn't allow at all.

paul90:
The following is from the perspective of a brief investigation into creating a version of Federated Wiki to make use of dat and beaker.

> We're wondering if that's too much of a pain for developers? The CSP limitations can be a nasty surprise.

[...]

PR #617, opened 2017-07-22
Relax the CSP for dat:// sites

pfrazee:
See the discussion in https://github.com/beakerbrowser/beaker/issues/552

This PR relaxes the CSP so that dat:// sites can access http/s origins. In a future PR, we will add an opt-in setting which defaults to the previous restrictive CSP, and then prompts the user to relax the CSP for a site that needs it.


pfrazee:
We modified the policy after to disallow HTTP embeds, because some folks were concerned about the security/privacy of HTTP

PR #619, opened 2017-07-23
Remove custom navigator.permissions API

pfrazee:
Now that we've relaxed the CSP (https://github.com/beakerbrowser/beaker/pull/617) there's no value in keeping this API. We're almost certainly going to replace this API with a different approach.


paul90:
But with 270063c6412ec6494866368a704d7f034bb27b8a referring to #617 is at best misleading. With the change in that commit the relaxation is only to https, and those that really do need to access to content from http sites are left out in the dark.

_Background: For historic reason 90% plus, of the existing Federated Wiki content is hosted on http only sites. While I would have hoped to have seen the number of https sites grow it has been very slow, and I expect some will never make that move._

For those parts of the federation not hosted with `dat` or `https` to be visible, we need to be able to get image and json files from the remote wiki sites. For this to work I believe that `default-src` would need to be extended to allow http. 

There is also a related CSP and CORS issue with hashbase, but I'll raise that elsewhere.

pfrazee:
@paul90 Sorry, I've not been explaining in detail. When we first made this change, we planned to add an opt-in setting where users can disallow HTTP/S by default (the old CSP) and then get automatically prompted to give permission when an HTTP/S request is attempted.

After relaxing the CSP, somebody pointed out the HTTP concern (as you know) so I tightened it back up again.

Now that we're not allowing HTTP by default for all users, we can use the same UI once its added: if a dat:// site attempts an HTTP request, it'll get blocked, and the user will be prompted to allow.

Automating it will be much nicer than this API, which required an explicit JS call. Nobody liked this API.

paul90:
Thanks @pfrazee - my comment is as much to ensure that sight is not lost of the need to automate this, to replace the API, as I don't see this on the on the milestones for 0.7.4 or 0.7.5

PR #623, opened 2017-07-25
Document default values for env vars

ralphtheninja:
I was wondering about the default values for storing data and such and couldn't find it in the readme so I thought I'd update it with the information I gathered.

Cheers


pfrazee:
Cool, thanks @ralphtheninja

PR #630, opened 2017-07-27
Use ModalActiveError

ralphtheninja:
https://github.com/beakerbrowser/beaker-error-constants/pull/1 has to land and be released before this can be merged.


pfrazee:
👍  thanks!

PR #633, opened 2017-08-03
Add beaker://debug-log

pfrazee:
This should help us debug some of the networking issues in dat.

screen shot 2017-08-03 at 1 23 29 pm
screen shot 2017-08-03 at 1 24 07 pm


pfrazee:
PRs welcome on either of those suggestions

PR #651, opened 2017-08-10
Dynamically Update Favicons

HughIsaacs2:
Fixes an issue where favicons wouldn't be able to be updated dynamically via JavaScript.
Beaker kept checking for the first favicon instead of the last, this brings it inline with other browsers.


pfrazee:
Ah cool, thanks 👍

PR #654, opened 2017-08-11
Add 'quit to tray' for persistent execution

### LINKED ISSUE #650: Run in background / minimize to tray mode
I've gotten a few requests to keep Beaker running in the tray after close, so that p2p sites stay available. Seems like a good idea to me. Anybody against?

pfrazee:
This PR updates Beaker so that it minimizes to the tray by default, rather than quitting. This keeps people's files online more reliably.

screen shot 2017-08-10 at 7 13 19 pm

Closes #650

### DIFF
--- a/app/background-process.js
+++ b/app/background-process.js
@@ -76,10 +76,3 @@ app.on('ready', function () {
   // listen OSX open-url event
   openURL.setup()
 })
-
-app.on('activate', () => windows.ensureOneWindowExists())
-app.on('open-url', (e, url) => openURL.open(url))
-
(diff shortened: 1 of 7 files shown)


pfrazee:
I need to test this on Linux before I land it.

pfrazee:
Also we want to do more with this -- current stats, maybe a fancier UI -- but that'll come later.

pfrazee:
Unfortunately linux requires an extra dependency to be installed via `apt` to support this. Going to just have it work for OSX for now.

PR #657, opened 2017-08-11
Run tests on travis-ci

G-Ray:
Add .travis.yml to run tests on travis-ci.

Tests are failing on linux and sometimes some tests fail on mac too, but at least we can have a CI set up for the p2p browser :+1:


pfrazee:
nice man, thanks 👍

PR #684, opened 2017-09-20
select the title input when opening the bookmark menu dropdown

frabrunelle:
I'm wondering if this could be a good change. Other browsers such as Chrome and Firefox seem to have this behavior (when you click to edit a bookmark, the title is automatically selected). It's a very minor change, but I've been thinking about it every time I open the bookmark menu dropdown, so I figured it might be a good idea to do a pull request 😛 

screen shot 2017-09-20 at 12 35 03 am

screen shot 2017-09-20 at 12 35 29 am


taravancil:
Thanks! That’s a great idea. I’ll merge in a bit when I’m at my computer

PR #687, opened 2017-09-22
add fullscreen options for youtube etc

mixmix:
I'm trying to use Beaker as my default browser. I was watching this excellent critique on masculinity in film :
 
[selection_400](https://www.youtube.com/watch?v=C4kuR1gyOeQ)*https://www.youtube.com/watch?v=C4kuR1gyOeQ*

But couldn't watch it fullscreen :crying_cat_face: _(blocked fullscreen / needy tabs + nav wanting to join the party)_

This PR aims to rectify that. Works well on Linux. Things that might still need attention : 
- what does this do on darwin?
- Hitting `Esc` while watching a video doesn't yet exit fullscreen mode, but it does get you your tabs + nav back.

### DIFF
--- a/app/shell-window/ui.js
+++ b/app/shell-window/ui.js
@@ -43,7 +43,7 @@ export function setup (cb) {
 
 function onWindowEvent (event, type) {
   switch (type) {
-    case 'blur': return docuument.body.classList.add('window-blurred')
+    case 'blur': return document.body.classList.add('window-blurred')
     case 'focus':
       document.body.classList.remove('window-blurred')
       try { pages.getActive().webviewEl.focus() } catch (e) {}
--- a/app/shell-window/ui.js
+++ b/app/shell-window/ui.js
@@ -42,11 +42,14 @@ export function setup (cb) {
 }
 
 function onWindowEvent (event, type) {
-  if (type == 'blur') { document.body.classList.add('window-blurred') }
-  if (type == 'focus') {
-    document.body.classList.remove('window-blurred')
-    try { pages.getActive().webviewEl.focus() } catch (e) {}
+  switch (type) {
+    case 'blur': return docuument.body.classList.add('window-blurred')
+    case 'focus':
+      document.body.classList.remove('window-blurred')
+      try { pages.getActive().webviewEl.focus() } catch (e) {}
+      break
+    case 'enter-full-screen': return document.body.classList.add('fullscreen')
+    case 'leave-full-screen': return document.body.classList.remove('fullscreen') 
   }
-  if (type == 'enter-full-screen') { document.body.classList.add('fullscreen') }
-  if (type == 'leave-full-screen') { document.body.classList.remove('fullscreen') }
 }
+
--- a/app/background-process/ui/window-menu.js
+++ b/app/background-process/ui/window-menu.js
@@ -194,11 +194,13 @@ export function buildWindowMenu (opts = {}) {
       label: 'Advanced Tools',
       submenu: [{
         label: 'Reload Shell-Window',
+        accelerator: 'CmdOrCtrl+alt+shift+R',
         click: function () {
           BrowserWindow.getFocusedWindow().webContents.reloadIgnoringCache()
         }
       }, {
         label: 'Toggle Shell-Window DevTools',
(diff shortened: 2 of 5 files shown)


app/shell-window/ui.js
```diff
@@ -42,11 +42,14 @@ export function setup (cb) {
 }
 
 function onWindowEvent (event, type) {
-  if (type == 'blur') { document.body.classList.add('window-blurred') }
-  if (type == 'focus') {
-    document.body.classList.remove('window-blurred')
-    try { pages.getActive().webviewEl.focus() } catch (e) {}
+  switch (type) {
+    case 'blur': return docuument.body.classList.add('window-blurred')
```
pfrazee: typo here

pfrazee:
Ok some other things we need to hit for this.

 1. [ ] Maximizing via mac's maximize button shouldnt hide the toolbar. That should only happen when the app requests fullscreen mode. So, we need to differentiate from fullscreen and maximized.
 2. [ ] When maximized but not fullscreen on mac, we need to render our own version of the traffic lights UI, because mac stops doing it for us. (Bleh.)
 3. [ ] We need to either ask permission before going fullscreen, or do what Chrome does and let the user know they've gone fullscreen. Otherwise it could be a way to phish users (go fullscreen, present fake browser UI, get user info).
    - If we want to ask first, that's as simple as going to `/app/lib/perms.js` and setting `alwaysAllow: false` on the fullscreen perm. The UX isn't as good though. Chrome's solution is definitely better.
 4. [ ] Open for debate (cc @taravancil), but I really like that chrome darkens the tab background when in maximized mode. I'd like to try that too!

I think that's everything. @mixmix let me know which of these todos are easy for you to tackle. The other ones you can leave for us.

app/shell-window/ui.js
mixmix: herp. Where my linting at! I'll use the script you've made next time (assuming that's the preferred pattern at the moment)

mixmix: ooooo, that linting setup is stale D:

mixmix:
okee @pfrazee , had a mull on this.

In short - I can't really dev the mac parts of this (I'm running linux on a thinkpad).

I have had a go at point (3) though. I've gone for **"tell the user they're entering fullscreen mode"** (and offer them the exit if they hover over the top border of the screen). 
[check it **here**](https://github.com/mixmix/beaker/pull/2) (more notes on a possible implementation there) 

I've done this instead of the permissions route because I would personally lose it if a browser asked me if I wanted to fullscreen _every time I clicked a fullscreen button_ :japanese_goblin: :fire:

app/shell-window/ui.js
pfrazee: Yeah I need to run it again and clean things up

pfrazee:
Great, I'll check it out

pfrazee:
Ok done. Thanks for making this PR @mixmix, I might not have gotten around to doing this otherwise. Check out what it ended up taking to get right: 3e5d9df1c97587c4f55403fd38cd0d941f9a9b1a

But it now works exactly like in Chrome!

mixmix:
Excellent. Sorry I couldn't follow this so the way through.
Interested to know how it was having this sort of contribution :
E.g. Did it pull you off task?

Certainly it's harder with intermittent contribution. Do you have an idea
of the threshold you want for adding someone as a contributor? That always
makes prs easier to modify in my experience

pfrazee:
> Excellent. Sorry I couldn't follow this so the way through.

No worries. TBH I can't imagine how you could've seen this one to the end. There is a *high* barrier of knowledge on this project, and I'm not sure how it could be lowered.

> Interested to know how it was having this sort of contribution: E.g. Did it pull you off task?

In this case, it got me to look at the problem again and see solutions that I hadn't seen the first time around. I don't know if it would always go as nicely, but this time it was a nice push. If it wasn't, I would've either requested the changes and left it, or explained why we couldn't handle the feature yet and closed the PR.

> Certainly it's harder with intermittent contribution. Do you have an idea of the threshold you want for adding someone as a contributor? That always makes prs easier to modify in my experience

I'll think about that. We haven't had enough outside PRs for it to come up, yet.

PR #688, opened 2017-09-24
add keyboard shortcust for shell window dev

mixmix:
I didn't know you could reload the shell-window without rebuilding :japanese_goblin: :sweat_drops: 

I guess I should have looked for or asked for advice on dev flow in this app.
Anyway, this adds shortcuts for accessing what I need to access right now.


taravancil:
Ah! We should add a note to the README about this. As an additional FYI, if the shell-window devtools are focused, you can reload the shell-window by doing the regular refresh keystroke

pfrazee:
Oh nice additions, thanks!

PR #717, opened 2017-10-21
Add onContextMenu to locationInput

rickycodes:
@pfrazee going after some low handing fruit...

Added a right click context so you can `cut`, `copy`, `paste` in `locationInput`:

contextmenu


pfrazee:
Nice 👍

PR #725, opened 2017-10-26
fix pageUp/Down shortcut

mixmix:
the original pull request had the direction inverted (compared to other browsers).
driving me crazy!


pfrazee:
Hah, thanks, I missed that

mixmix:
phew, everything is right in the world again.

Issue #671, opened 2017-08-29
Typing in URL bar sometimes interrupted


taravancil:
Sometimes when I'm typing in the URL bar, the URL input loses focus and interrupts my typing. I haven't figured out what triggers this, but it happens to me multiple times a day.

Does anyone else experience this?

millette:
Related: #679 (for reference).

taravancil:
Hi @Butterfly, this happens when I have multiple tabs open/loading, and then click `Cmd+T` to open a new tab and start typing. In Beaker (and every other browser afaik), opening a new tab automatically focuses the URL input. But in Beaker, this bug interrupts the element being focused while I'm typing.

The nice thing about it,  is that's it's defintely a bug. Typing in the URL input should *not* be interrupted. We'll fix it before the 0.8 release 😸

PR #727, opened 2017-10-27
Fix URL typing interruption

alvesjtiago:
Fix issue reported by @taravancil (https://github.com/beakerbrowser/beaker/issues/671) that prevented URL from keeping active when other pages finished loading.


taravancil:
I’m at mozfest so I won’t be able test this out for a bit, but thank you!!

alvesjtiago:
@taravancil no problem at all. Let me know once you get a chance if it completely fixes the issue you were experiencing 😊

taravancil:
Thanks for this! It seems to have fixed the issue.

PR #730, opened 2017-10-28
Add Windows binary & Update build instructions

whfeeds:
I updated the readme to match to https://beakerbrowser.com/docs/install/#building-from-source

thanks!


pfrazee:
Thanks! :+1:

PR #744, opened 2017-11-05
Add back, forward and reload menuItems

rickycodes:
adding some familiar menu items:

menuitems


pfrazee:
Thanks for the contribution!

I'm 👍  this except that it needs to gray out forward/backward if it can't go in that direction.

You can use the `webContents` methods: https://electron.atom.io/docs/api/web-contents/#contentscangoback

rickycodes:
@pfrazee good call!

Updated (and 100% better)

pfrazee:
Nice work! Merging

Issue #679, opened 2017-09-03
Feature request for URL bar autocomplete


taravancil:
Can we get this?

<img width="212" alt="screen shot 2017-09-03 at 1 55 12 pm" src="https://user-images.githubusercontent.com/7584833/30005780-9321e9ce-90af-11e7-929f-d1106ab48f45.png">

If I type "twi" in other browsers, they'll pre-populate the URL input, so I can simply click enter to go to twitter.com. This trips me up regularly in Beaker. That flow is burned into my muscle memory.

SaFrMo:
@taravancil @pfrazee I'm looking to contribute to Beaker and am wondering if this is a good place to start. Do you have any thoughts on how you'd like to see this implemented? If not, I'll get started on my own. Thank you!

millette:
Is there already a Last Recently Used concept in Beaker?

webdesserts:
@millette From what I saw Beaker does have autocomplete suggestions, but you have to tab down to them right now.

@taravancil @pfrazee Although I started looking into this, I will be busy with family & work for the rest of the week. So if @SaFrMo wants to take a stab at it that's fine with me 👍. 

@SaFrMo From what I observed in Firefox, this feature could be broken down into the following requirements:

1. Auto-fill the rest of the URL bar when one of the auto-suggestions starts with the currently typed string.
2. Select the untyped characters of the auto-suggested URL.
3. Highlight the corresponding auto-suggestion in the list so that when you hit `Enter` it will launch that suggestion
4. Ensure that Backspace won't trigger auto-filling the url.

SaFrMo:
Thanks @webdesserts for the suggestions! Looks good to me - I'll start work on it in the next day or two 👍

pfrazee:
FYI that URL bar is pretty complex, so I hope that's not an overly discouraging element to work on. There are a couple of code smells in the chrome UI, and that's one of them (though it's not only incidental complexity at fault).

SaFrMo:
[Snapshot](https://github.com/SaFrMo/beaker/commit/ee46bda5de01d4ddfb3340f53ce50a0525c9094a) of current work - so far, I've got two changes going:

1. The selection of the first matching autocomplete is working, and 
2. The new `autocompleteSuggestion` var keeps track of the characters to add to the end of the current input value - so the `er.com/` in this issue's original screenshot.

pfrazee:
@SaFrMo cool! looking forward to seeing the final product

[...]

PR #803, opened 2017-12-30
URL bar autocomplete

SaFrMo:
Written to solve #679. To reproduce:

1. Start typing a URL of a previously visited page.
1. Beaker will find the most-visited page starting with the value you entered, autocomplete the URL, and select the relevant result in the autocomplete dropdown.
1. Hit "Backspace" to delete the autocomplete and return to the top dropdown result.


pfrazee:
I just spent 5 minutes trying to break this UI element and it worked perfectly! Very clean code too. I'm very impressed, nice work @SaFrMo. A+ would merge again.

PR #814, opened 2018-01-12
Render swarm debugger error in view

samuelmaddock:
I added an error message to the view when parsing the URL fails. Normally the view renders a blank screen when this error occurs.

screen shot 2018-01-11 at 8 31 43 pm


pfrazee:
👍  Thanks

PR #840, opened 2018-02-07
[WIP] Restore Window State on Crash

### LINKED ISSUE #737: Feature request: Restore tabs from last time
Operation System: OSX 10.11.6
Beaker Version: 0.7.8

Wishing for a setting to bring back the tabs from the last session when the browser is restarted.

I'm one of those users who keeps twenty tabs open to read later.  Losing my tab state is a major usability issue.

This would also make crashes less painful.

webdesserts:
**Do Not Merge** This is a work-in-progress pull request to fix #737. Right now this branch just always restores the previous window state. I still need to make sure that this only happens on crash. Would also probably want to bring up a prompt to ask if they want to restore state. I know some browsers also have a "Restore tabs from last time" setting to have this always trigger, so I'm not sure if I should try to sneak that into this pull request as well. Mainly I want to get the code in front of y'all to see if you have any feedback on the approach. Just want to see if I'm on the right track here. Please feel free to ignore any `console.log` statements I'll try to remove those before all's said and done.

window-state-example

### DIFF
--- a/app/background-process/open-url.js
+++ b/app/background-process/open-url.js
@@ -5,7 +5,7 @@ var queue = []
 var isLoaded = false
 
 export function setup () {
-  ipcMain.on('shell-window-ready', function (e) {
+  ipcMain.on('shell-window:ready', function (e) {
     queue.forEach(url => e.sender.send('command', 'file:new-tab', url))
     queue.length = 0
     isLoaded = true
--- a/app/background-process/ui/windows.js
+++ b/app/background-process/ui/windows.js
@@ -7,6 +7,7 @@ import path from 'path'
 import * as openURL from '../open-url'
 import * as downloads from './downloads'
 import * as permissions from './permissions'
+import {debounce} from '../../lib/functions'
 
 const IS_WIN = process.platform === 'win32'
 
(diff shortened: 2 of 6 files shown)


pfrazee:
Cool this is coming along well. If we can reliably detect crashes, a "restore?" dialog sounds like a good idea. We can also add a menu item under View or History to restore last session.

webdesserts:
@pfrazee what's a good place to store the flag for "the last session crashed"? Another thought here is that we could write the crash log to a file somewhere (if that's not already done somewhere) and then check for the existence of that file at startup.

pfrazee:
@webdesserts actually what I'd probably try first is to store a flag indicating that there was a clean exit. I'd put it in the `shell-window-state.json` file, and make sure to unset it on every start. If you load and the file exists but the flag is unset, you know a crash occurred and you need to restore.

webdesserts:
@pfrazee makes sense. I'll try it out this evening and let you know how it goes.

webdesserts:
@pfrazee Ok, I've implemented the crash detection bit. Try it out and let me know what you think (I still want to go back and clean up my code).

You can uncomment [this line] to test a crash.

[this line]: https://github.com/beakerbrowser/beaker/pull/840/files#diff-4b0a8a5cc61e7d34c4f6d679249dbdc8R74

pfrazee:
Great work! I made a few updates and merged. Most of my updates were organizational and formatting. The only logic change was, I updated `defaultBrowsingSessionState()` to have a `true` clean exit. That way first-run doesn't accidentally get detected as a crash. There should be no risk of a crash in the middle of saving session state losing the file entirely, because you used fs-jetpack's [atomic write](https://stackoverflow.com/questions/17047994/transactionally-writing-files-in-node-js) mechanism.

PR #852, opened 2018-02-14
Implemented mouse browser back/forward navigation (Windows only :c)

DTV96Calibre:
This PR adds and registers a new event handler function for handling App Commands emitted by a Windows OS. Only the 'browser-backward' and 'browser-forward' commands are being handled by this code. When either are handled, the appropriate calls are made to navigate forward or backward in browsing history.


pfrazee:
👍  Looks good!

PR #857, opened 2018-02-18
[WIP] Add an "About Beaker Browser" window

zeke:
screen shot 2018-02-18 at 12 55 13 pm

I was reporting a beaker issue yesterday and went looking for my app version in the "about" menu. Alas, no such menu was to be found! This PR changes that, using the [electron-about-window](https://ghub.io/electron-about-window) package created by Electron superhero @rhysd.

To Do:

- [ ] Figure out where the PNG icon file should actually live (@pfrazee? @taravancil?)... am I correct in assuming that only the contents of the `app` dir are available at runtime in the packaged app?
- [ ] Windows
- [ ] Linux


pmario:
There is some info at: beaker://settings/#general  and beaker://settings/#information
Just to let you know.

Your dialog is nicer!

Jacalz:
This looks wonderful :+1:

PR #858, opened 2018-02-18
add a keyboard shortcut to preferences on mac

zeke:
Open preferences with <kbd>Cmd</kbd><kbd>,</kbd> on macOS

screen shot 2018-02-18 at 1 07 20 pm


pfrazee:
👍  Looks good, thanks!

PR #867, opened 2018-03-15
Fix running beakerbrowser outside the repo

mafintosh:
After this fix you can run `beakerbrowser` anywhere on your machine to launch it after npm linking


pfrazee:
Thanks!

Issue #145, opened 2016-10-18
Dockable devtools


pfrazee:
Currently, devtools are only available as a popped-out window. It's a pain in the ass. They ought to be dockable. Related issue: https://github.com/electron/electron/issues/3177

millette:
Just use a tiling window manager ;-)

taravancil:
Good news! This will be possible in Electron 2 https://github.com/electron/electron/pull/11300

PR #912, opened 2018-03-23
Dockable devtools (WIP)

pfrazee:
Closes https://github.com/beakerbrowser/beaker/issues/145

Todos:

 - [x] Merge electron v2 update (separate issue, should happen first)
 - [ ] Make detachable
 - [ ] Make closable by X button
 - [ ] Make resizable

### DIFF
--- a/app/background-process/adblocker.js
+++ b/app/background-process/adblocker.js
@@ -7,13 +7,13 @@ export function setup () {
   initialize() // contains-ads
 
   const beakerUrls =  /^(beaker|blob)/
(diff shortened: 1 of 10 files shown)


swati-maheshwari:
@pfrazee  How? does it work for webview?

pfrazee:
@swati-maheshwari The solution we used is to switch to the `BrowserView` API instead of webviews, which work pretty differently. It basically comes automatically with browser views

swati-maheshwari:
@pfrazee is there any solution for webview? to open docked debugtools

pfrazee:
@swati-maheshwari sorry, I never solved that

PR #913, opened 2018-03-24
Support tab navigation in navbar autocomplete

brechtcs:
Other browsers allow to move up and down in their navbar autocomplete using Tab and Shift-Tab. This patch adds support for this in Beaker.


brechtpm:
Not exactly sure why this is failing, but at first sight it doesn't look related to my changes.

pfrazee:
Hey, I dont know if we're going to add this or not. Let's discuss post 1.0 release

PR #936, opened 2018-04-04
MIT License isn't modified

benwiley4000:
Looks like the license was un-modified in https://github.com/beakerbrowser/beaker/commit/6c9304d0bfab15e77c6801ed21bd7a927e065bbf.


pfrazee:
Yep, thanks

PR #949, opened 2018-04-13
Implement find-next and find-previous shortcuts

msfeldstein:
menu and keyboard shortcuts for find next (cmd+G) / previous (Shift+cmd+G). Remove the dependency on the click event, i assume there will only ever be one Find window in the ui so we can look it up directly.


pfrazee:
👍 merged! Thanks!

PR #954, opened 2018-04-16
Replace workspaces with a bidirectional syncing folder (WIP)

pfrazee:
One of the chief UX struggles has been finding a tool which lets developers build Dat sites while simultaneously controlling what they publish.

Anything written to a Dat can be downloaded by other users. Since dev involves making lots of little changes, we didn't want to pollute the history with each change.

**Old solution**

The "staging area" and "workspaces" were an attempt to solve that. They created a sort of *draft* version. The draft version would read directly from a folder, giving the user the chance to do their dev-work. Then, when ready, the user could publish to the dat.

This worked okay, but it had a problem once other apps started writing to dats with the `DatArchive` API. Basically it forced other apps to be aware of the draft. Sometimes this meant doing extra work, like writing `commit()` after a set of writes, which was a pain. Other times the app *wanted* access to the draft, but didn't have enough tools to handle them well. It felt like both the "staging area" and "workspaces" were inviting too much complexity.

**Re-approaching the problem**

We stepped back and decided to write out our requirements:

 1. Users need a draft version to do their development
 2. Users need to control what's in the final "published" version
 3. Apps sometimes need to write/read the draft version
 4. Apps sometimes need to ignore the draft version and write directly to the "published" version. (Related: sometimes a draft version isnt needed.)
 5. Users sometimes want automatic bidirectional sync of dats to a folder, whether they're working on it or just saving/viewing their data.
 6. Users sometimes want to share their draft versions with other users

**New solution**

With these requirements in mind, we came up with a new idea which this PR starts. This is the "syncing folder" approach. It works like this:

 - Get rid of workspaces and the staging area.
 - Every dat can have a "syncing folder." UX-wise this will be similar to setting the workspace folder; you do it in the library during a new dat's setup.
 - The syncing folder will bi-directionally sync the dat with its folder. The user can change the files and the changes will automatically publish.

This new solution solves requirements 1, 3, 5, and 6, but it needs a solution for handling the "published" version. To do that:

 - Dats will be able to diff and merge between each other.
 - The user will be able to set a dat as the "publish target" of another dat.
 - As a result, the user will get a diffing UI that's similar to the current workspaces UI, but it will be from the "draft dat" to the "published dat" instead of from a workspace folder to the published dat.
 - (Possibly) there will be metadata indicating when a dat is a draft or the publish target.

This new approach will hopefully eliminate a lot of the complexity. It's dats all the way down. Users can easily setup the automatic bidrectional sync. The publish target is then an optional extra step. When apps want to differentiate between where they're publishing, they just have the user tell them which dat to write to (the draft or the published version). Hopefully this will be much clearer to users as a result.

This PR will only include the syncing folder. The publish target controls need a little more thought, and so we'll handle them in a subsequent PR.


pfrazee:
Important note --

This update creates a new DB field for the syncing folder and does not try to migrate over your old paths.

... kind of sucks but it was the only safe path, since the behaviors are changed from workspaces.

pfrazee:
This PR also introduces a "default .datignore" setting so that you can prepopulate the datignore on all your projects. (Important to help us avoid accidentally publishing node_modules all the time.)

pfrazee:
👍 We're moving forward with this

paul90:
> This PR also introduces a "default .datignore" setting so that you can prepopulate the datignore on all your projects. (Important to help us avoid accidentally publishing node_modules all the time.)

Might also be an idea to include those OS specific files, like `.DS_Store` or `Thumbs.db`, that we will always want to avoid publishing. Though this might be better done with a global datignore

webdesserts:
@pfrazee Heads up, I tried this out and it seems that if you sync with an existing folder, some of the files in that folder can be inexplicably wiped out. I think it might be that the contents of a folder that does not yet exist in the dat are wiped. I'll try to reproduce later.

paul90:
> * Dats will be able to diff and merge between each other.
> * The user will be able to set a dat as the "publish target" of another dat.
> * As a result, the user will get a diffing UI that's similar to the current workspaces UI, but it will be from the "draft dat" to the "published dat" instead of from a workspace folder to the published dat.
> * (Possibly) there will be metadata indicating when a dat is a draft or the publish target.

For a *draft* dat the user should be able to set a "publication target", any linking back from the "published" dat to a "draft" must be private. When we get multi-writer support this provides a hint on where we look for changes that need reflecting back into our draft, via the diffing UI.

Having metadata to indicate a *draft* dat, together with something in the UI to help prevent it being inadvertently shared would be nice. That is not to say, with multi-writer support, that having the draft on multiple devices should not be possible.

It would be desirable to support having a more restrictive `.datignore` on the published dat. An example might be to prevent sharing notes, or files associated with an editing/development environment, that we want to keep with the "draft" dat.

pfrazee:
@webdesserts huh ok I'll beef up my automated tests around this. I ran a fair amount of tests myself; sorry that I missed some cases.

@paul90 good thoughts. Agree about the draft dat staying private and having UI help. Differing .datignores would make sense.

Issue #961, opened 2018-04-22
JSON Rendering


DakshMiglani:
Operation System: Windows 
Beaker version: Latest

Add: JSON Rendering just like markdown rendering.

PR #962, opened 2018-04-22
Add JSON Rendering

DakshMiglani:
# Added JSON Rendering.

for issue: #961

Dependencies added: `json-formatter-js`

Possible Use Cases? : 

1. Beautifies the json
2. Helps developers understand json
3. ...addMore

### DIFF
--- a/app/json-renderer.js
+++ b/app/json-renderer.js
@@ -1,7 +1,7 @@
 import createJSON from './lib/fg/json'
 
 const isJSON = function (str) {
-  if (str.blank()) return false
+  if (str === "" || str === ' ') return false
   str = str.replace(/\\(?:["\\\/bfnrt]|u[0-9a-fA-F]{4})/g, '@')
   str = str.replace(/"[^"\\\n\r]*"|true|false|null|-?\d+(?:\.\d*)?(?:[eE][+\-]?\d+)?/g, ']')
   str = str.replace(/(?:^|:|,)(?:\s*\[)+/g, '')
--- a/app/json-renderer.js
+++ b/app/json-renderer.js
@@ -11,8 +11,7 @@ const isJSON = function (str) {
 if (!document.querySelector('main')) {
   var unformattedEl = document.querySelector('body > pre')
   if (isJSON(unformattedEl.textContent)) {
-    var formattedEl = document.createElement('main')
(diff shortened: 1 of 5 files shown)


DakshMiglani:
@pfrazee can you check this?

taravancil:
Hi @DakshMiglani! We saw this but haven't reviewed it yet! Be patient - we have 100 or so other things on our TODO list :smile:

DakshMiglani:
@taravancil Oh, alright. paul didn't reply so i thought you guys have not got the time. It's all cool tho.

app/json-renderer.js
```diff
@@ -0,0 +1,21 @@
+import createJSON from './lib/fg/json'
+
+const isJSON = function (str) {
+  if (str === "" || str === ' ') return false
+  str = str.replace(/\\(?:["\\\/bfnrt]|u[0-9a-fA-F]{4})/g, '@')
+  str = str.replace(/"[^"\\\n\r]*"|true|false|null|-?\d+(?:\.\d*)?(?:[eE][+\-]?\d+)?/g, ']')
+  str = str.replace(/(?:^|:|,)(?:\s*\[)+/g, '')
+  return (/^[\],:{}\s]*$/).test(str)
```
pfrazee: Would it be simpler to just do `JSON.parse(str)` and return false if it throws?

DakshMiglani: let's say there's a very large JSON string, parsing it to the memory might even cause the browser to crash just to know that it's json, instead just use a regular expression to understand if it is of type json or not.

pfrazee: That's true. In that case, I'd suggest we check the string length first, and return false if it's >2mb. It'd make sense not to handle huge JSONs anyway, right?

DakshMiglani: yea that can be done.

DakshMiglani: @pfrazee can you put the exact bounds for the max length of the string, i can submit a few commits.

pfrazee:
Merged! Thanks for the contribution!

pfrazee:
We ended up changing the theme to be light, and adding a raw/formatted toggle interface

DakshMiglani:
Oh, thanks!

PR #971, opened 2018-04-30
Add a shortcut for focusing the address bar (alt + d).

Happy0:
When using beaker browser, I find myself missing the alt + d shortcut to focus the address bar (which is available in firefox and chrome.)

While investigating what the situation is on Mac, I noticed there's an alternative 'ctrl + l' and that beaker implements that here: https://github.com/beakerbrowser/beaker/blob/bbed013686ddf0f8d78c9afbc8e4dbc3da3780d9/app/background-process/ui/window-menu.js#L108

(Oh, if I hold 'alt' I can see it labelled within the file menu.)

If you are interested in having the 'alt + d' shortcut as well, would you prefer that I instead either:

* Make it possible to define an accelerator of the form 'CmdOrCtrl+LOrAlt+d'
* Make it possible to define an array of accelerators in the menus ?

Or is this pull request fine as is?

### DIFF
--- a/app/background-process/ui/windows.js
+++ b/app/background-process/ui/windows.js
@@ -141,6 +141,7 @@ export function createShellWindow (windowState) {
   registerShortcut(win, 'Ctrl+PageDown', onNextTab(win))
   registerShortcut(win, 'CmdOrCtrl+[', onGoBack(win))
   registerShortcut(win, 'CmdOrCtrl+]', onGoForward(win))
+  registerShortcut(win, 'alt+d', onFocusLocation(win))
 
   // register event handlers
   win.on('browser-backward', onGoBack(win))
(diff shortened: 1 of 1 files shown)


Happy0:
Oh, I see there's a precedent for having alternative shortcuts in `window.js` for things that also have shortcuts in the menus, so perhaps my approach is fine?

pfrazee:
Hey @Happy0. We're 👍 with adding alt+d as another "focus location" shortcut. I think this probably is the best way to do it, since electron menu-items can't accept multiple accelerators. Merging!

Happy0:
\o/ yay.

Does that make me a beaker browser developer now? :3

pfrazee:
Hah sure :)

PR #983, opened 2018-05-09
Library (list) UI tweaks

pfrazee:
Removes the "Recent" grid and instead...

 1. Makes the listing larger and show more info
 2. Adds sortable columns

I liked the recent grid, but the full listing (beneath) wasn't nice to use, and we couldn't expand the grid view to cover all items.

This PR uses just one tabular view that's got larger fonts, larger spacing, and more overall information.

## Before

screen shot 2018-05-08 at 7 45 49 pm

## After

screen shot 2018-05-08 at 7 44 19 pm


pmario:
I like it! .. But I would love to have 2 more settings. 

 - A tag ... which would allow different archives to be fitered and sorted eg: "A-B test"
 - a color ... Just to get more visual difference

At the beginning we have just a view archives. ... But I'm sure I'll have 10++ ;)

eg: 

beaker-whishlist

pmario:
I personally could live without the size info.

pmario:
Tag and Color should be independent. ... sorting by color imo isn't needed.

pmario:
tag and color should not be published. For me it's local only.

pfrazee:
@pmario hey sorry for the delay - I think the tag and color ideas are pretty interesting and I'll be thinking about them, but I'm going to move slowly while we think through all the issues with management

pmario:
I do like the latest changes. Also saw your intro video. gr8 stuff :)

pfrazee:
:) thanks!

taravancil:
Cool! I added some commits. Should be good to go now.

PR #986, opened 2018-05-11
'Dedicated peer' integration and services DB (WIP)

pfrazee:
Big update! This PR:

 1. Adds a services DB for managing accounts with [PSA-Compliant](https://github.com/beakerbrowser/beaker/wiki/PSA-Web-Service-Discovery-Protocol) services, specifically services that support [DEP-0003: HTTP Pinning Service API](https://www.datprotocol.com/deps/0003-http-pinning-service-api/). That includes Hashbase and Homebase.
 2. Adds `beaker.services` internal Web API for managing the services DB and exchanging with the service-accounts.
 3. Adds a UI to the library-view page for viewing current pins on the users' "dedicated peer" (aka pinning) services.
 4. Adds a UI to the settings page for managing services.

Screenshot of the current state of 3:

screen shot 2018-05-11 at 6 13 59 pm

This PR is not ready yet. @taravancil needs to do a pass on all the UI work.

### DIFF
--- a/app/background-process.js
+++ b/app/background-process.js
@@ -26,6 +26,7 @@ import * as settings from './background-process/dbs/settings'
 import * as sitedata from './background-process/dbs/sitedata'
 import * as profileDataDb from './background-process/dbs/profile-data-db'
 import * as bookmarksDb from './background-process/dbs/bookmarks'
+import * as servicesDb from './background-process/dbs/services'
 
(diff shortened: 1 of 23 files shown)


knownasilya:
Could the PSA service be an intent? Where by default its Hashbase?

paul90:
It might be nice to add some visual feedback to show when the 'dedicated peer' is sync'ed up.

PR #988, opened 2018-05-15
Add compare tool to library-view

pfrazee:
This tool lets you diff between Dats and publish/revert files between them.

screen shot 2018-05-14 at 7 19 15 pm
screen shot 2018-05-14 at 7 19 22 pm
screen shot 2018-05-14 at 7 19 30 pm
screen shot 2018-05-14 at 7 19 38 pm


pmario:
Looks good. I'll test soon and report back.

pmario:
If I need to scroll down in compare mode, I can't see, which direction "publish" or "revert" actuall is. .. So do I copy from A -> B or B -> A

I would like the tooltip to be more descriptive. eg: `test -> new` for [publish]  AND `new -> test` for [revert]

beaker-lib-diff-01

pmario:
Version 17 and latest are differnt! It would be nice see the differences between 2 versions, to resurrect deleted content if needed. 

beaker-lib-diff-02

pmario:
The [swap] function is a very nice feature and extremely useful ... but ... the [swap] button should be more visible.

webdesserts:
@pfrazee @taravancil Looks good 👍🏼 I think its a good idea to start here. Provide developers with something simple and wait to see where the missing use cases are. My only feedback is that the "publish" feature is a bit hidden.

I'm wondering if it will be immediately apparent to devs who aren't following this conversation that `workspaces -> publish` is now hidden behind `compare -> select archive -> publish`. If we are taking Github as an example, they have a call to action on the landing page of a repository in the form of a **New Pull Request** button. You don't have to go digging through navigation to know it's an option:

screenshot 2018-05-15 17 26 53

If we aren't going to provide a "draft" solution in the immediate future, it might be good to at least provide some hints to devs on how they might use "Compare" to replace the old Workspaces workflow. For example, when they land on an empty "Compare" page, we could use that as an opportunity to say teach the user how to make a makeshift draft:

> _Select an archive_ to compare with _Test Website_. You can review the changes and merge them together.
>
> **Hint:** Want to draft some changes, but aren't ready for the world to see them yet? Make a copy of your site and use this page to "publish" the changes when you're ready!

I'm mostly viewing this from the perspective of a user who has never seen Github, but knows they don't want to "save their changes" yet. In their case, "Compare" might not be self-explanatory. This is a bit of a worst-case scenario, but hey we already have kids creating sites in Beaker so I don't think it's too unlikely. This was kind of my thought behind ultimately implementing a draft process, nevertheless it would be interesting to see if we can find a way to survive without.

pfrazee:
Thanks for the feedback @pmario and @webdesserts.

@webdesserts I'm thinking that, for the moment, "Compare" will be a slightly more advanced feature. We can find ways to direct people to it, and we can document it, but my thinking is that automatic-publish is suitable for the average user. (The lack of drafts is reflective of our limited resources -- trying to get best possible without holding up 0.8 much longer.)

Ideally what we will figure out is a way for somebody to think, "How do I choose when things get published?" and then they can find that answer quickly.

In the meantime, we can spend some time trying this tool and thinking about it.

taravancil:
Let's sit on this before merging. There are still a number of unanswered questions that I don't have time to answer right now. Listing them here as TODOs for myself:

[ ] - The language around publishing and reverting isn't clear. What's being published to what? What gets reverted?
[ ] - In the dropdown for selecting the target, Untitled archives are blank. I think we should also included the shortened dat URL next to the title so dats are easier to identify.
[ ] - I don't love the arrow icon, let's replace that

taravancil:
We’re just getting back into the swing of things after some international
travel. I won’t give you a timeline, but yes, eventually there will be
progress.

pfrazee:
We're going to merge this into `draft-mode-rebased` and update it to work with the new drafts system

PR #991, opened 2018-05-16
Exclude unused native binaries from builds

samuelmaddock:
Release builds are including all platform binaries for sodium-native which adds a lot to the build size. This change includes only binaries for the current platform.

```bash
# Before change
~/projects/beaker (master) $ du -sh ./dist/mac/Beaker\ Browser.app/
240M    ./dist/mac/Beaker Browser.app/

# After change
~/projects/beaker (master) $ du -sh ./dist/mac/Beaker\ Browser.app/
225M    ./dist/mac/Beaker Browser.app/
```

It looks like Beaker only uses sodium-native within the background process so only the node binaries are required. I wasn't sure which ABI version is needed so it still includes all node binary versions for the platform.

Note that I've only tested this on macOS.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,12 @@
   },
   "build": {
     "appId": "com.pfrazee.beaker-browser",
+    "files": [
+      "**/*",
+      "!node_modules/**/sodium-native/prebuilds/**/*",
+      "node_modules/**/sodium-native/prebuilds/${platform}-${arch}/libsodium.*",
+      "node_modules/**/sodium-native/prebuilds/${platform}-${arch}/node-*"
+    ],
     "asar": true,
     "asarUnpack": [
       "./node_modules/sodium-native/**"
--- a/package.json
+++ b/package.json
@@ -30,13 +30,16 @@
     "appId": "com.pfrazee.beaker-browser",
     "files": [
       "**/*",
-      "!node_modules/**/sodium-native/prebuilds/**/*",
-      "node_modules/**/sodium-native/prebuilds/${platform}-${arch}/libsodium.*",
-      "node_modules/**/sodium-native/prebuilds/${platform}-${arch}/node-*"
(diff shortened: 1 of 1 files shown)


mafintosh:
Ya that looks right to me.

pmario:
@samuelmaddock .. how does your build command look like? I can test ubuntu 18.04 and windows 10

samuelmaddock:
@pmario I followed the steps in the readme to produce a build: https://github.com/beakerbrowser/beaker#building-from-source

pmario:
TLDR; I'm not sure, if it is worth it!!

----

I did a test with windows 10 ... It compiles and the app seems to work ... but

On my machine the existing installer contains this: 

beaker-test-bins-02

After the patch is looks like this: 

beaker-test-bins-01


That's ok for my machine, but it will miss the 32bit versions. ... 

If that's OK, then the patch seems to work. .. I didn't run the tests!!!

The size savings are about 15 Mbyte on windows. Installed size is about 242 MByte (optimzed verison)

samuelmaddock:
@pmario The configuration uses [file macros](https://www.electron.build/file-patterns#file-macros) to target the configured platform and architecture for each build. I don't think Beaker Browser is configured to build for 32-bit Windows. However, if it was, it would be a separate build only containing the 32-bit sodium-native binaries.

samuelmaddock:
@mafintosh @pfrazee Update: I fixed this up so it works with `npm run release` and it also now strips out unused utp-native binaries. Here's the size update with the latest changes on master.

```
~/projects/beaker (master) $ du -sh ./dist/mac/Beaker\ Browser.app/
215M    ./dist/mac/Beaker Browser.app/
~/projects/beaker (remove-unused-bins) $ du -sh ./dist/mac/Beaker\ Browser.app/
183M    ./dist/mac/Beaker Browser.app/
```

32MB (15%) size savings! 🎉

Issue #984, opened 2018-05-10
Cmd + A in raw JSON view should only select the JSON


taravancil:
Cmd + A currently selects the JSON and the Formatted/Raw toggle buttons

PR #1024, opened 2018-06-04
Don't select json files UI

msfeldstein:
per #984, dont let the "Formatted/Raw" buttons be selectable, so when you cmd+a you just get the content.


pfrazee:
@msfeldstein Thanks! I ended up doing basically this but using the CSS rules. Closed!

PR #1037, opened 2018-06-09
Add DatArchive#checkout() method

pfrazee:
Currently, if you want to access a specific revision of an archive, you have to manipulate the URL. This snippet is an example of what this is like:

```js
// read from latest
let vi = archive.url.indexOf('+')
if (vi !== -1) {
  archive = new DatArchive(archive.url.slice(0, vi))
}
```

This PR adds a `checkout()` method to the `DatArchive` which makes it easy to do this. If no number is passed in, you will get latest. Examples:

```js
var v5 = archive.checkout(5) // get revision 5
var latest = archive.checkout() // get latest
await archive.checkout(5).readdir('/') // read the files listing at root of revision 5
```

### DIFF
--- a/app/builtin-pages/com/archive-history.js
+++ b/app/builtin-pages/com/archive-history.js
@@ -15,12 +15,9 @@ export default function render (archive) {
 
   // lazy-load history
   if (archive) {
-    // strip the version
-    let vi = archive.url.indexOf('+')
-    if (vi !== -1) {
-      archive = new DatArchive(archive.url.slice(0, vi))
-    }
-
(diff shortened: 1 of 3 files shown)


pfrazee:
Anybody object to this addition to the `DatArchive` API?

When more sophisticated version controls (like checkpoints) are added, this method will have to be expanded to support them

joehand:
sweet!

RangerMauve:
LGTM!

pfrazee:
Response has been good so I'm merging

Issue #1023, opened 2018-06-02
Show percentage of data "given back" to the network


k-nut:
Trying to put into writing here what I briefly discussed with @taravancil at the Berlin p2p-event.

The cool thing about Beaker is that by visiting websites you not only consume them but you can also start giving back by seeding them. The way that you are giving back is not very visible yet right now. Other p2p apps such as transmission (see screenshot) show a ratio of the amount downloaded vs the amount uploaded. I think it would be cool to show something like this on a per website basis with beaker, too.

I also think that showing a ratio specifically could be interesting because at the point where you have reached a ratio of 1 you are effectively "giving back" more than you are "taking" which I think can be quite an empowering feeling.

I'd be happy to look into this if you agree that this might be a good idea and have any ideas for where I could start. 

![screen shot 2018-06-02 at 08 12 38](https://user-images.githubusercontent.com/1096357/40871503-084c484e-663d-11e8-9f5d-dac93d0ed170.png)

pfrazee:
Yeah I think this is a cool idea. We'll probably want to track a lot of different stats around each dat's resource usage at some point, so that we can set policies to cap usage. This is something we could include in that.

k-nut:
Glad you like it! Do you have any pointers for where one would start with this? Is this something to be implemented on the beaker level or can you see that there will be changes required in some other library that beaker depends on?

pfrazee:
Yeah I'm happy to describe what I'm thinking if you want to try tackling it.

 - In the dat/library.js, I'd probably add logic for tracking stats on each archive. I'd draw from the [dat-node stats code](https://github.com/datproject/dat-node/blob/master/lib/stats.js) and probably track the transfer rate per archive (not per connection) as well as the total bytes transferred. (Other stats? Not sure.)
 - In the DatArchive Web API implementation of `getInfo()`, I'd put those stats inside the object provided to `beaker://` pages.
 - I'd put those stats somewhere in the network view of the "library-view" builtin page.

pfrazee:
I can also imagine putting those stats in the shell-window's dropdown that shows how many peers there are.

taravancil:
I think you might have closed the wrong issue

pfrazee:
Shoot, yep

todrobbins:
👍 👍 👍

k-nut:
@pfrazee cool, I think I found a way that I can track those stats, essentially by doing something like  `archive.content.on('upload', updateCounterForArchive)`.
[...]

PR #1044, opened 2018-06-15
add per archive and session stats about uploads

k-nut:
As discussed in #1023.

This is just a very first approach, I'm not sure if attaching the stats to the `archive` object is a good idea and the texts should probably be rephrased but this can be taken as a first proof of concept 🙂

### DIFF
--- a/app/background-process/networks/dat/library.js
+++ b/app/background-process/networks/dat/library.js
@@ -10,6 +10,7 @@ import through from 'through2'
 import split from 'split2'
 import concat from 'concat-stream'
 import CircularAppendFile from 'circular-append-file'
+import networkSpeed from 'hyperdrive-network-speed'
 var debug = require('debug')('dat')
 import * as siteData from '../../dbs/sitedata'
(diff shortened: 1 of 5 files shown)


pfrazee:
Good start! I'm in the process of refactoring the codebase, and tara's working on the library-view UI -- so when both of those settle, I'll pull this code in.

k-nut:
hey @pfrazee 👋, seeing that this PR is in conflict with master now, should I try to fix those conflicts or are there still other things that you want to finish first before pulling this in?

pfrazee:
@k-nut thank you! But no need. I'm going to pull this in (soon) but the prerequisite for adopting this PR is for me to do a rethink of the network tab. And-- I'm going to be doing that pretty soon! So I think we'll get stats like these involved then.

pfrazee:
Okay, this information is now in the network tab of the dat source-code view (what beaker calls the "library"). I'm going to close, but we may choose to add some of these stats to the navbar UI in the future.

Thanks!
screen shot 2018-09-03 at 3 36 35 pm

PR #1047, opened 2018-06-17
Donate dropdown overflows with long URLs

HughIsaacs2:
This isn't really a complete fix, it'll still look a bit wild on screen but it should keep the URL from escaping the dropdown for the Donate button (I tested the feature with a dat:// URL to come across this bug).

Could've used "word-break: break-word" or "overflow: hidden" instead but that's a design thing and I'm not sure I've good input to share there.

### DIFF
--- a/app/stylesheets/shell-window/donate-menu-dropdown.less
+++ b/app/stylesheets/shell-window/donate-menu-dropdown.less
@@ -27,5 +27,6 @@
   .body {
     margin: 0 10px;
     padding: 10px 0;
+    overflow-wrap: break-word;
   }
 }


HughIsaacs2:
donate-bug
Image for clarity.

HughIsaacs2:
donate-bug
Screenshot for clarity.

pfrazee:
Appreciate it! Beyond the overflow issue, I'm not totally happy with this UI yet. I feel like we need something more attention-grabbing, but not annoying. Maybe words next to the icon? (" ❤️ Support this site"). Maybe a horizontal bar prompt that auto-opens on page load, like this?:

image

Basically, @taravancil needs to take a pass at it and work her magic.

HughIsaacs2:
I think words next to it is a good idea. Especially since the heart today is associated with the like button on many websites and even sometimes with bookmarking.

Also I think if it auto-opens there should be some criteria similar to how Chrome on Android handles PWAs, where it opens after a few uses of the site, basically saying "Hey, we've noticed you've visited this site quite a bit, why not show support by visiting the donation page?".
Auto opening might be a problem though, since it could be associated with the annoying permission prompts many sites are just shoveling out today (geolocation, notification, etc...).

What if after several views of a site, the heart jumped or something, like the icons in the dock on macOS?
Something attention grabbing, but small and inviting.

pfrazee:
> What if after several views of a site, the heart jumped or something, like the icons in the dock on macOS?

Yeah that's not a bad idea, something like that. We'll think about it.

Thanks for the fix!

Issue #1033, opened 2018-06-07
Allow user to change search engine


sskyy:
Operation System: MacOS
Beaker version: 0.8.0-prerelease.6

I could not find a place to change search engine, and I find that duckduckGO is hard coded in the program. I would be happy to give a PR if you approve.

pfrazee:
Yeah this has been on my todo list. I'd be happy to hit it myself, or to accept a PR.

PR #1052, opened 2018-06-19
[WIP]: start adding support for other search engines. ( issue #1033 )

Happy0:


Happy0:
Hullo @pfrazee . I seen you say you'd maybe accept a pull request for this on #1033 so I thought I'd give it a shot to get familiar with the code base =].

Not sure if I'm going in the right direction here or not...

My main question for the moment is - is it possible to get at the settings somehow from the ` app/shell-window/ui/navbar.js` file? I see that the 'beaker' global (where you can get the settings from) isn't available here - I assume this is deliberate?

pfrazee:
Hi @Happy0 !

Let's aim to have a few builtin options (DuckDuckGo, Google, Yahoo, Bing) and then an open input for the user to choose. For the open input, I think we'll need to store at least 2 data-points:

 - Name / Label
 - Query URL

Where "Query URL" is a template similar to `https://duckduckgo.com/?q=%s`. Since the user will be picking from a few presets and then from their own custom choice, I think we'll actually need 3 data-points in the settings:

 - `search_engine`: One of the predefined engines, or `"custom"` if the user inputs their own.
 - `search_engine_name`: The name/label of the engine
 - `search_engine_url`: The URL template of the engine

These values will be set in all cases, so the predefined versions will simply give fixed values for `_name` and `_url`.

When used, we'll read `search_engine_url` and replace `%s` with the search query.

 > My main question for the moment is - is it possible to get at the settings somehow from the app/shell-window/ui/navbar.js file? I see that the 'beaker' global (where you can get the settings from) isn't available here - I assume this is deliberate?

It should be available, but perhaps not declared in the globals at the top of the file.

aral:
Unlike mainstream browsers and given the social mission of Beaker, it is irresponsible to provide surveillance capitalists like Google, Yahoo, and Bing – whose business models are based on tracking and profiling people – without at least issuing a warning. Otherwise, Beaker would be complicit in exposing the people using it to surveillance. Also, we should be visibly separating them from DuckDuckGo whose business model is _not_ based on violating your privacy.

pfrazee:
I'd be +1 to having DuckDuckGo as the only builtin option, and have users manually enter their alternatives

Happy0:
Yeah, let's do that :). I was a bit squeamish about having Google, etc as
selectable options as well. I'll update my PR to just be DuckDuckGo and a
customised option.

Would you mind if I added some autocomplete search functionality (
http://easyautocomplete.com/example/duckduckgo-ajax-api ) for DuckDuckGo
while I'm at it? Or would you rather I did it in a separate pull request?

When using a custom Search engine I guess we'd just not autocomplete the
search as the user types in the address bar / search widget?

aral:
Sounds great – thanks both! :)

Also, for future reference – in case we want some other convenient alternatives – here are some other search engines that do not track:

  * Findx: https://www.findx.com (has own crawler/doesn’t use Google under the hood)
  * Qwant: https://www.qwant.com
  * StartPage: https://www.startpage.com

pfrazee:
@taravancil do you have an opinion on adding duckduckgo autocomplete? Y/N?

taravancil:
I just checked out that [example](http://easyautocomplete.com/example/duckduckgo-ajax-api) - it could be a nice enhancement, but I think it should live in a separate PR :)

I'm also not quite sure about how I feel about that flavor of autocomplete in the URL bar. @Happy0 do you know if there are any browsers that support it? I want to get a sense for how it would feel in Beaker before moving forward.

aral:
@taravancil It does seem more useful than the Safari autocomplete (see screenshot) – but I guess it comes down to how accurate the entires are. PS. The example seems to have some edge cases (see “undefined” entries in second screenshot). CC @pfrazee 

screen shot 2018-06-26 at 10 10 10

screen shot 2018-06-26 at 10 10 56

pfrazee:
I think the autocomplete could be cool- let's try it in a separate PR

Happy0:
Thanks all :). 

I should be able to work on this again (and possibly autocomplete) on Thursday.

PR #1059, opened 2018-06-28
fix(archive-history): Use an anchor tag for archive-history-items

knownasilya:
The current implementation uses a div with `onclick` and `href` without `role='button'` (or equivalent). I think this is a more correct implementation of that. Let me know if it makes sense. I only tested this in devtools, since I can't build beaker ATM (see freenode channel).

Not sure if overriding the default `a` styles is the best way, maybe there is already code to do this better?

### DIFF
--- a/app/builtin-pages/com/archive-history.js
+++ b/app/builtin-pages/com/archive-history.js
@@ -30,9 +30,7 @@ export default function render (archive) {
         // render
         var rowEls = history.map(c => (
           yo`
-            <div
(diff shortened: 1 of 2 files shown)


taravancil:
Hey Ilya! Thanks for this and sorry for the trouble you've had building Beaker. We'll see if we can get it figured out in IRC.

I'm going to close this without merging because we're doing some substantial changes in another branch that have touched this component, so changes here will likely result in merge conflicts 😿

Issue #1068, opened 2018-07-06
Rename Finder to Explorer on Windows.


jraspiprojects:
Operation System: Windows 10
Beaker version: Latest

You need to change Open in Finder to Open in explorer on windows, because we don't have finder, only explorer.exe

pfrazee:
Good idea, added to our TODOs

jraspiprojects:
Where would "Open in Finder" or the downloads page be in the source?

jraspiprojects:
Just do 
`if (process.platform == "win32"){
 actions = [ actions = [
          yo`
            <span class="link show" onclick=${e => { e.stopPropagation(); downloadsList.showDownload(d) }}>
              Show in Explorer
            </span>
]} else {actions = [
          yo`
            <span class="link show" onclick=${e => { e.stopPropagation(); downloadsList.showDownload(d) }}>
              Show in Finder
            </span>
]}`

jraspiprojects:
Hmm, can't get code formatting to work correctly

jraspiprojects:
@pfrazee I made a PR, feel free to look at it and fix any formatting errors...

pfrazee:
Fixed in master

PR #1076, opened 2018-07-13
Update downloads-file.js

jraspiprojects:
I am not too fluent in javascript and node so feel free to edit it @pfrazee 

(Pull request for #1068 )


pfrazee:
@jraspiprojects I appreciate it! But I think we can actually get a simpler form of this. See https://github.com/beakerbrowser/beaker/commit/82212a4a48300d39be2297867e7d8dd02cb5eaf4. Less duplicated code that way.

Issue #1072, opened 2018-07-10
Add a setting to control where files are downloaded to


pfrazee:

jraspiprojects:
Where is the code for the downloading of files? I can't see it anywhere.

pfrazee:
https://github.com/beakerbrowser/beaker/blob/master/app/background-process/ui/downloads.js

jraspiprojects:
I created a PR

Issue #1073, opened 2018-07-10
Ask permission before downloading a file


pfrazee:
We currently auto-download files, and we should prompt the user first

webdesserts:
@pfrazee just so I'm understanding this right, is this referring to files downloaded to the "Downloads" page/folder?

pfrazee:
Yeah

PR #1085, opened 2018-07-19
Update downloads.js

jraspiprojects:
For #1072


pfrazee:
I actually just landed some code to get permission for the download (see https://github.com/beakerbrowser/beaker/issues/1073). #1072 is about choosing the default directory which downloads go to.

PR #1110, opened 2018-08-05
Make restore session dialog more understandable

matt1:
"Restore session" vs "Cancel" is confusing if you are not familiar with the software - it felt like it was going to cancel loading the app.  This makes it clearer what the choices are I think.


pfrazee:
Oh very nice, yes thank you.

Issue #1104, opened 2018-08-02
Downloads show "Open in Finder" on Windows


matt1:
Operation System: Windows 10
Beaker version: Version: 0.8.0-prerelease.7 Electron: 2.0.5 - Chromium: 61.0.3163.100 - Node: 8.9.3

Completed downloads show a "Show in Finder" option.
Expected to say "Show in Explorer" on Windows.

I guess from looking at the code, it will say "Show in Finder" on linux too.

matt1:
It appears that the process.platform variable is undefined when in the context of rendering the template, so this would always fall through to "Show in Finder" here https://github.com/beakerbrowser/beaker/blob/master/app/builtin-pages/com/downloads-list.js#L68

pfrazee:
Ah. Thanks. Will fix.

jraspiprojects:
Done.

jraspiprojects:
And on 64-Bit Windows machines, it returns win32.

matt1:
I believe that win32 is expected even on win64 (this is an node thing).

Might be worth leaving this bug open to make sure that Linux users also get some attention (otherwise even with this PR, they'd still get "Show in Finder" rather than something more generic like "Show in File Manager")

jraspiprojects:
Actually we could replace that whole if statement with Show in File Manager for all!

taravancil:
I’d like to keep the per-OS specificity

pmario:
I'd vote for a generic button, that fits all OSes. 

Distinguishing between "Finder" and "Explorer" is not enough. For Ubuntu it's "Files" ... 

Using the OS App search for
 - "finder" shows nothing, 
 - "explorer" shows the browser. 
 - "files" shows the app, that's meant. 

Other distros may use different names for their default "File Manager" __and__ The default can be changed.

pfrazee:
I agree with @pmario. I changed it to just say "Show in folder".

jraspiprojects:
lol

PR #1111, opened 2018-08-06
Look at #1104

jraspiprojects:
#1104


app/builtin-pages/com/downloads-list.js
```diff
@@ -61,13 +61,18 @@ export function render (downloadsList) {
             <i class="fa fa-times"></i>
           </button>
         `
-
+        if (process.platform == "win32"){
```
matt1: I am guessing you've checked this, but just to be sure the "real" node process is in scope here right?  i.e. it has the process.platform property?

Also a small nit: probably safer to use === instead of ==

jraspiprojects: true

pfrazee:
We're just going to use generic language. Thanks though!

PR #1210, opened 2018-10-12
[WIP] Watchlist Feature and Built-in Page

Septimus:
Beginnings of the watchlist feature. 
1. Created a built-in page for the watchlist.
2. From the built-in you can add sites, see which have been resolved or not, and choose to seed them when they become available.
3. Added a notification badge to the dropdown menu that will tell you if there are any resolved sites since the last time you looked.

[WIP]
Still need to test several aspects to the watch list and there are a few things that I'd like to improve or expand on before complete.

### DIFF
--- a/app/background-process/protocols/beaker.js
+++ b/app/background-process/protocols/beaker.js
@@ -287,6 +287,15 @@ async function beakerProtocol (request, respond) {
   if (requestUrl === 'beaker://settings/main.js') {
     return cb(200, 'OK', 'application/javascript; charset=utf-8', path.join(__dirname, 'builtin-pages/build/settings.build.js'))
   }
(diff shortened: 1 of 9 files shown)


pfrazee:
Solid work on this. I made a few adjustments:

1) I moved the watchlist notification into the browser UI itself so it's a little easier to notice

screen shot 2018-10-13 at 4 58 56 pm

2) I updated some aesthetics on the watchlist page itself

screen shot 2018-10-13 at 4 58 44 pm

3) I decided to remove the "seed after resolved" for a couple reasons, but the main one was that it added complexity that I just don't think we need that badly.

4) You did a good job with the adder-menu on the top right but I'm going to turn it into a modal

Issue #1214, opened 2018-10-17
Saving files limited to 100 kb. 


DanielRJansen:
Operation System: macOS 10.13.6
Beaker version:0.8.0

Imported new TiddlyWiki (empty.html) opens and works. In Library it Shows:
 
390 lines ( actually 11218 lines )
2.08 mb  ( accurate )

When edited and saved:

390 lines ( all lines after 390 missing )
100 kb ( accurate )

I tried an HTML full of lorum ipsum...

330 kb, 12000 lines, saved to 100.003 kb and 875 lines.

Septimus:
I can definitely confirm that the lines are getting cut, but I am not seeing a loss in file size. I created a file with ~2500 lines equaling 702.47kB. It is now down to ~339 lines but still ~702kB after save.

Will look into the issue on lines cutting off, and decreasing file size should I end up seeing that myself.

Septimus:
Also seeing the loss of file size. I guess it takes a moment to update that.

pfrazee:
Fixed, thanks @DanielRJansen for reporting and @Septimus for fixing this.

PR #1217, opened 2018-10-18
Fix to #1214

Septimus:
Renders a preview max `1e5` (100,000 bytes) in size. When you click the pencil and enter edit mode, the full file will be shown in the editor. Not showing on initial viewing to prevent rendering a large file.


pfrazee:
👍 Great work, thanks @Septimus !

Issue #937, opened 2018-04-04
Feature: Switch by default to P2P version of web site when available


aaronshaf:

todrobbins:
👍 👍 👍

pfrazee:
Okay here's how I propose we do this:

 - [ ] Add a global setting to "Automatically redirect to dat:// when available", default on (`auto_redirect_to_dat`)
 - When a dat:// version is detected...
   - [ ] ...If `auto_redirect_to_dat=1`, then immediately redirect to the dat:// version
   - [ ] ...else, show the current "P2P version available" badge
 - [ ] After an auto-redirect to dat://, show a new badge that says "HTTPS version available" which will redirect to the https:// version. When clicked, the auto-redirect-to-dat for that site should be disabled for the browsing session.

pfrazee:
Feature merged! Thanks to @Septimus for doing this!

PR #1218, opened 2018-10-19
Default to dat feature #937

Septimus:


app/shell-window/pages.js
```diff
@@ -32,6 +32,8 @@ export const DEFAULT_URL = 'beaker://start'
 
 export const APP_PATH = remote.app.getAppPath() // NOTE: this is a sync op
 
+export var noRedirectOrigins = new Set()
```
pfrazee: Can you add a comment explaining what this is?

app/shell-window/pages.js
```diff
@@ -137,6 +139,7 @@ export function create (opts) {
     liveReloadEvents: false, // live-reload event stream
     zoom: 0, // what's the current zoom level?
     retainedScrollY: 0, // what was the scroll position of the page before navigating away?
+    autoRedirectToDat: null,
```
pfrazee: Same, needs a comment

app/shell-window/pages.js
```diff
@@ -149,6 +152,7 @@ export function create (opts) {
     sitePerms: null, // saved permissions for the current page
     siteInfoOverride: null, // explicit overrides on the siteinfo, used by beaker: pages
     siteHasDatAlternative: false, // is there a dat:// version we can redirect to?
+    siteHttpAlternative: null,
```
pfrazee: Same, needs a comment

app/shell-window/pages.js
```diff
@@ -203,10 +207,24 @@ export function create (opts) {
...
 
     // wrap webview loadURL to set the `loadingURL`
-    loadURL (url, opts) {
+    async loadURL (url, opts) {
+      // setting up for auto redirect
+      let redirectToDat = await beaker.browser.getSetting('auto_redirect_to_dat')
+      let hasDatAlternative = false
+      if (!url.includes('dat://')) {
+        hasDatAlternative = await DatArchive.resolveName(url).catch(err => false).then(res => !!res)
+      }
```
pfrazee: This implementation looks good but `loadURL()` isn't called in every case where we need this behavior to trigger. The place it *must* run is `checkForDatAlternative()`, because that will be called reliably on every load.

Putting it in `loadURL()` will be somewhat faster because the check is done *before* we load the target. So I think it makes sense to have it in both methods.

app/shell-window/pages.js
```diff
@@ -203,10 +207,24 @@ export function create (opts) {
...
+
       // reset some state
       page.isReceivingAssets = false
       page.siteInfoOverride = null
+      page.autoRedirectToDat = !!redirectToDat
+
+      if (page.autoRedirectToDat && hasDatAlternative && !noRedirectOrigins.has(url)) {
+        page.siteHttpAlternative = parseURL(url).protocol
+        console.log(page.siteHttpAlternative)
```
pfrazee: Dont forget to remove this log()

app/shell-window/ui/navbar.js
```diff
@@ -679,6 +691,19 @@ function onClickGotoDatVersion (e) {
   }
 }
 
+function onClickGotoHttpVersion (e) {
+  const page = getEventPage(e)
+  if (!page || !page.protocolInfo) return
+
+  const url = `https://${page.protocolInfo.hostname}${page.protocolInfo.pathname}`
+  pages.noRedirectOrigins.add(url)
```
pfrazee: We should do these by origin or hostname (hostname might be easier). If we're disabling the dat redirect, we really want to disable for all pages on that location.

Septimus: Agreed.

pfrazee:
👍 Great work!

I made a few updates (c780701c31a7055d10a876cfe4f5466a78f435ca, 1f8fdd5a3f7f6a0bc7fd1329c2151dfb3140a3c0) and merged. There was a bug in the `checkForDatAlternative()` function that caused redirects to fail (the `url` param was being imported from the closure and wasn't correct) so I updated the function to also take the full URL.

PR #1253, opened 2018-11-07
Quick fix to library search

Septimus:
Allow dat:// to be apart of the dat key search.
Now works with just keys or dat://key


app/builtin-pages/views/library.js
```diff
@@ -98,11 +98,12 @@ async function loadArchives () {
 function filterArchives () {
   if (query && query.length) {
     archives = archives.filter(a => {
+      console.log(a)
```
pfrazee: Leftover log

pfrazee:
Remove the leftover log and we're 👍

pfrazee:
👍 👍 👍

PR #1278, opened 2018-12-07
Beaker editor

Septimus:
Note: FontAwesome 5 icon issues will need to be fixed for this PR.

**Beaker Editor V1**

- Open dat archive in the editor from the library using the context menu
- Support for preview mode with a diff editor
- File explorer
- Ability to have multiple files open using a tab system
- Use cmd/ctrl + s to save

v1 contains the core of the editor more features to come!


app/builtin-pages/com/editor/file-tree.js
```diff
@@ -0,0 +1,279 @@
+const yo = require('yo-yo')
+import * as models from './models'
```
pfrazee: Good work on this. +1 for building file-tree by copying from existing code; that was the right call.

I'm going to suggest we make a change to its structure.

Unless you plan for there to be multiple `FileTree` instances, let's change this from a class to just a regular module. We can just move any state contained by the class into a `// globals` section. Any methods can become exported API functions.

I suggest this because it's currently halfway between a module and a class, and I think it'll be clearer if we just go 100% module.

app/stylesheets/photon.min.less
```diff
@@ -30,7 +30,7 @@ input[type=number]::-webkit-inner-spin-button,input[type=number]::-webkit-outer-
 input[type=search]::-webkit-search-cancel-button,input[type=search]::-webkit-search-decoration{-webkit-appearance:none}
 fieldset{border:1px solid silver;margin:0 2px;padding:.35em .625em .75em}
 legend{border:0;padding:0}
-*{cursor:default;-webkit-box-sizing:border-box;box-sizing:border-box}
+*{-webkit-box-sizing:border-box;box-sizing:border-box}
```
pfrazee: Omfg thank you. Why did we not do that sooner.

pfrazee: (Photon is a hangover from a really early iteration and it's basically been an angry monkey on our back for...forever.)

pfrazee:
Okay, things are looking good. A few notes from my tests:

 - Readonly dats need to indicate that they're not editable. Currently they behave as usual, and then error when you attempt to save.
 - On a site with a preview, I'm not seeing anything in the "Preview changes" area. I can investigate why if you need.
 - If we want to release the MVP without the bottom bar controls, let's hide it for the release.

I have more that we can discuss on a call.

PR #1298, opened 2019-01-06
test(datpeers): add ability to get own peer id

imsnif:
I'm including a test here for https://github.com/beakerbrowser/beaker-core/pull/18
I could not find any unit tests in the @beaker/core repo, so I assumed the integration tests here are sufficient. :)

I chose to create the `enableExperimentalDatPeersInDatJson` method from an existing test so that it won't depend on same test to run. Am definitely open to suggestions if this is not desirable.


pfrazee:
Yeah looks good 👍

PR #1328, opened 2019-01-29
More Analytics: Count Dat archives and test global net reachability

da2x:
Extend analytics to track these additional usage metrics:

*netReachablePort*
Asks a remote server to test whether the client’s Dat port is reachable or not. Using Transmission’s test server as a proof of concept until this can be moved to Beaker’s own analytics server. Data will help determine how many people are capable of acting as nodes and participating in the network.

*datWatchlist*
Number of items on the watchlist. Data will help prove how useful this feature is and also how many people regularly have problems downloading Dat archives.

*datArchivesOwned*
Number of archives the client has created. Data will help establish whether people are creating with Beaker.

*datArchivesSeed*
Number of archives the client is seeding. Data will help determine whether people actively choose to volunteer bandwidth and help keep the distributed network healthy.

*datArchivesCache*
Number of cached archives. Data will help determine whether people are engaging with content on the distributed web.

I’ll work on the server side next if there is interest in adding support for the above. I don’t believe the above constitutes personal data nor do I believe any user would object more to participating in analytics data collection with or without the above data points.

### DIFF
--- a/app/background-process/analytics.js
+++ b/app/background-process/analytics.js
@@ -9,7 +9,7 @@ import jetpack from 'fs-jetpack'
 import concat from 'concat-stream'
 import osName from 'os-name'
 const settingsDb = beakerCore.dbs.settings
-import {ANALYTICS_DATA_FILE, ANALYTICS_SERVER, ANALYTICS_CHECKIN_INTERVAL} from '@beaker/core/lib/const'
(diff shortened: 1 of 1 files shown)


pfrazee:
Sorry this never got discussed, we were focused on the 1.0 release. We'll do a pass on analytics in the near future

Issue #1246, opened 2018-11-05
Enable UPnP port-forwarding


da2x:
Beaker is supposed to be for normal users that don’t fiddle with their router settings, right? Then it should pick a random TCP port number for Dat and setup port-forwarding using UPnP for that port by default.

This should help increase the number of reachable seeders. It should also give people a better experience when they experiment with peer-to-peer distribution in Beaker.

Note that Beaker shouldn’t use a hard-coded port number with UPnP in case there are more than one client on the same network. Assigning a random port number for every session is also the preferred solution anyway as some [crappy budget] routers have issues with renewing/reassigning the same port number and will return a failed state for existing port mappings even when they point to the same internal IP that is requesting reassignment.

Using a hard-coded port number with IPv6 and falling back to one on IPv4 when UPnP fail is in line with what other software does.

RangerMauve:
Beaker is going to be using [hyperswarm](https://github.com/hyperswarm/network) for all it's networking, so it might be a good idea to check to see how this would work there.

nettiopsu:
@RangerMauve @pfrazee btw, do you have some approximate timeline when is it gonna be released with hyperswarm?

pfrazee:
Unfortunately no. Maf is workin hard on it.

PR #1340, opened 2019-02-16
UPnP support

da2x:
This is a partial fix for #1246. This will only work for one  device behind the same NAT as all clients wants to  control the same port rather than using normal random port assignment. This would have to retry and negotiate with the router in addition to tracking which port that should be announced to the tracker.

I implemented this in Beaker rather than Dat as this required the least amount of change and will work as a temporary solution while we wait for Hyperswarm. This commit should be reverted when Hyperswarm is done.

I’ve evaluated all the UPnP NPM modules I could find and none of them are all that great. They silently don’t handle errors and don’t negotiate IGDv1/v2 properly. There are two common ways for port forwarding (permanent and ttl-leased) usually depending on the available memory on the router. I’ve tested it on two different consumer routers., but I believe this implementation should work for both permanent and ttl-leased port assignments for one device behind most edge routers.

### DIFF
--- a/app/background-process.js
+++ b/app/background-process.js
@@ -14,6 +14,7 @@ import * as rpc from 'pauls-electron-rpc'
 import * as beakerBrowser from './background-process/browser'
 import * as adblocker from './background-process/adblocker'
 import * as analytics from './background-process/analytics'
+import * as portForwarder from './background-process/nat-port-forwarder'
 
 import * as windows from './background-process/ui/windows'
 import * as modals from './background-process/ui/modals'
@@ -79,6 +80,8 @@ app.on('ready', async function () {
   // start the daemon process
   var datDaemonProcess = await childProcesses.spawn('dat-daemon', './dat-daemon.js')
 
+  portForwarder.setup()
+
   // setup core
   await beakerCore.setup({
     // paths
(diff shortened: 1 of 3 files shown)


da2x:
Here is some data on why this is needed: I’ve monitored peer announcements for dat://8bb99a0ac0d56131e61bd4bfea74ced9bc2c603e78e4eb72e8dd2fd6f33743cd (https://beakerbrowser.com) on discovery1.datprotocol.com over the last 48 hours. I’ve seen 76 different IP/port addresses and only a single one of them have been reachable; an address on Google Cloud that I’m assuming is the seeder at Hashbase.io.

pfrazee:
👍 Code looks good. Going to merge and see how it goes. Thanks!

RangerMauve:
Any chance this could be added to discovery-swarm instead?

pfrazee:
@RangerMauve Would be fine with me. Is discovery-swarm being used outside the node stack at all? That's the only consideration I can think of

RangerMauve:
Not that I've heard of yet. I think that the people using discovery-swarm on mobile for instance, are using it with `nodejs-mobile`. Not sure if this would be too much harder to compile in that context.

PR #1373, opened 2019-04-03
Handle relative donation URLs in location bar

### LINKED ISSUE #1048: Donate link feature doesn't support relative URLs
donate-error

If a Dat site has a relative URL set as the Donate link, clicking it sends the user to beaker://shell-window/{the URL}.
So setting it to something like /donate/ sends the user to [beaker://shell-window/donate/](beaker://shell-window/donate/).

figueredo:
Closes #1048.


pfrazee:
👍 Thanks!

Issue #1318, opened 2019-01-08
CSS: user-select:none on menu's items


DaniellMesquita:
![image](https://user-images.githubusercontent.com/6765277/50865581-71148f00-138d-11e9-92d3-73b04a224df2.png)

PR #1375, opened 2019-04-03
user-select: none for issue #1318

HughIsaacs2:
Added user-select: none on the .menu-item class (A fix for issue #1318).


pfrazee:
@HughIsaacs2 Thanks for the PR! Did you test this? I'm 99% your change is to dead code that I haven't removed yet.

HughIsaacs2:
Honestly untested.

Just felt it'd be a safe bet considering it'd only disable highlighting on the menu items themselves and not the whole menu.

pfrazee:
Okay thanks

PR #1380, opened 2019-04-06
[WIP] Re-enable testing on Linux

da2x:
So what exactly breaks?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ env:
 
 os:
   - osx
-#  - linux
+  - linux
 language: node_js
 node_js: "8"
 


pfrazee:
The tests have actually improved quite a bit since linux was removed, but they still don't work perfectly in travis. Currently the `experimental-dat-peers-web-api-test` fails (see https://travis-ci.org/beakerbrowser/beaker/jobs/516547348#L3698 -- all tests announce themselves to help us figure out which one is breaking in cases like this)

pfrazee:
Automated tests have been removed for now. We'll need to readd them with the new beaker 1.0

PR #1411, opened 2019-06-01
Nudge people towards the forum when filing bugs

da2x:
The [Dat forum](https://dat.discourse.group/) is open for registration and can help keep support requests and discussions out of the bug tracker. It also provides better search and forum tools.


pfrazee:
👍 !

PR #1453, opened 2019-08-18
feature: enable adblocker using @cliqz/adblocker-electron

remusao:
Hi there,

As a follow-up to this tweet: https://twitter.com/cblgh/status/1163102926707089408 I thought it would be interesting to see if adblocking can be enabled in Beaker again using [@cliqz/adblocker-electron](https://www.npmjs.com/package/@cliqz/adblocker-electron). I have put together a small proof-of-concept and it seems to work fine locally.

The way it currently works is this:

1. When `adblocker.setup()` is called, the resources for the adblocker are fetched (<2 MB of rules used to block ads: easylist, etc.) and an instance of `ElectronBlocker` is created.
2. Blocking is enabled in `session.defaultSession` by registering listeners to `webRequest.onHeadersReceived` and `webRequest.onBeforeRequest`, as well as registering a preload script which allows to further hide and block ads which cannot be removed via network filtering alone.

Optionally, the instance of `ElectronBlocker` can be serialized (to a typed array `Uint8Array`) and persisted on disk for faster retrieval. This can be done from `adblocker.js` or from a `postinstall` script which would create this file at build-time instead of runtime (none of this is implemented in the pull request yet).

Last but not least, currently the setup is only targeted at blocking ads, but not tracking. This means that filters such as `easyprivacy` are not enabled. The result is, less resources are needed to perform blocking, and there should be less breakage in practice (although this is usually very small even when anti-tracking rules are enabled).

Let me know what you think about it, happy to make some changes if you would like to move forward with this.

Best,
Rémi

### DIFF
--- a/app/background-process/adblocker.js
+++ b/app/background-process/adblocker.js
@@ -1,24 +1,30 @@
 import { session } from 'electron'
-// import { initialize, containsAds } from 'contains-ads'
+import { ElectronBlocker } from '@cliqz/adblocker-electron'
+import fetch from 'node-fetch'
 
 // exported API
 
 export function setup () {
-  /*
-  temporarily disabled due to https://github.com/bbondy/bloom-filter-cpp/issues/7
-    "contains-ads": "^0.2.5",
-
-  initialize() // contains-ads
-
-  const beakerUrls = /^(beaker|blob)/
-  session.defaultSession.webRequest.onBeforeRequest(['*://*./*'], (details, callback) => {
-    const shouldBeBlocked = !details.url.match(beakerUrls) && containsAds(details.url)
-
-    if (shouldBeBlocked) {
-      callback({cancel: true})
-    } else {
-      callback({cancel: false})
-    }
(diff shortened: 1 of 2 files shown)


remusao:
I am unsure of the cause of failing tests, could someone more knowledgeable give me a pointer as to what might be the issue? I saw that in the previous adblocker implementation, an exception was made for requests containing the following patterns:

```javascript
const beakerUrls = /^(beaker|blob)/
```

Should this be maybe re-introduced?

pfrazee:
Hey, so sorry for the late reply. I'm +1. If you don't mind, I'd like to let this PR sit until I'm finalizing the 0.9 updates, and then I'll roll it in then. Sounds good?

remusao:
Hey, no worries at all. This plan sounds great! Do let me know if I can help in any way. Happy to update the PR after 0.9 to get it in shape.

pfrazee:
Excellent, thank you for doing this!

pfrazee:
Merged a separate PR but ❤️ to @remusao for doing this.

PR #1505, opened 2020-02-07
made webterm use keycodes to detect special/control keys

karroffel:
This patch changes the history/input handling code of the
webterminal to use keycodes instead of the name of keys.

Using direct keycodes means that different keyboard layouts or
custom input methods will work properly.

The code of the key-mapping object is taken from the `mousetrap`
library which is licensed under the Apache-2.0 license.


pfrazee:
Tested, works great! Thanks!!

PR #1541, opened 2020-05-01
Minor change

frabrunelle:
Considering that this context menu also shows up when you right-click on a folder, I think it could make sense to rename it to Export. This is also how it's called in the Files Explorer. I also renamed it to Export for the actions menu just to be consistent but could also make sense to leave it like that.


pfrazee:
LGTM!

PR #1729, opened 2020-06-08
added 'Save Link As' functionality to context menu on right clicking URL 

### LINKED ISSUE #1619: Right-click Save As option for downloading files from links
If I want to download a file from a link in Beaker (ex: a PDF file), I need to open the link then save the file using the "File... Export Page As" menu option.  It would be nice to have the Save As or Export function on the context menu for the link.

Example: hyper://0179c63933ac650471d69d3ae70a1b1755407b70a8a7b98cd21f47b8e670bc9d/

matthewauld:
Fixes #1619 . First pull request ever on an open source, please let me know if I should be doing anything differently. Right click on url now displays "Save Link As" functionality. If there is no file extension, defaults to html.


pfrazee:
Thanks @matthewauld! I decided to change the order so I fixed it directly, but otherwise this was a great PR and I appreciate you doing the actual work for me!

Issue #1669, opened 2020-05-24
Can’t open about:blank


da2x:
**Describe the bug**
about:blank triggers a search instead of opening an empty HTML document.

**To Reproduce**
1. Click on the address field and type "about:blank" and press Enter

**Expected behavior**
Should open a blank website. (This is in the HTML spec. Part of the window object, I think.)

**Actual behavior**
It triggers a web search instead.

**Desktop (please complete the following information):**
- OS: Linux, macOS, Windows
- Browser: Beaker
- Version: 1.0-pr2

pfrazee:
I'll look into this

PR #1735, opened 2020-06-10
about URI's accepted as valid as opposed to triggering search

matthewauld:
fixes issue #1669. Another massive pull request! `about:blank` redirects correctly, but all other required `about:` uri's redirect to `about:blank#blocked`. I am not sure if that is chromium issue or how important it is, however, this is a step in the right direction.


pfrazee:
👍  Thanks!

Issue #1722, opened 2020-06-05
Change the default search engine


tiotrom:
Is it possible to change the default search engine? DuckDuckGo is full of ads and in Beaker you can't install any extenssions to block those ads. I recommend any SearX instance instead of Duck - https://searx.space/

pfrazee:
Not currently, if anybody wants to implement this, it needs a setting (in the db and the settings page) and then the location-bar will need to be updated to pull from that choice.

matthewauld:
Yeah, these are all things you would implement on top of beaker browser rather than built into it. A search engine for beaker would be a great project!

pfrazee:
Distributed search is something we're pretty interested in once Hypercore APIs land

rafapaezbas:
Hello I am looking for some contribution I could work on and this one seems pretty good for me. Anyway I have observed that default search engine can already be changed. Is this a feature you are still interested on? Any suggestions for a useful PR on this topic? Thanks in advance!

PR #1737, opened 2020-06-10
Added ability to select search engine

matthewauld:
#1722  Added the ability to select from a list (right now just google.com and duckduckgo.com,) or your own, custom search url.

A few notes:
- in the settings menu, the customs fields don't do any sort of validation. You need to enter a valid url. I am not sure if there is a standard URL regex we are using or not - I can add that in.
- urls.js `examineLocationInput` returns a vSearch result, which I changed just to return the query string to search (eg. `?q=query1+query2`) as opposed to the full url. It is only used in the location bar code as far as I can tell, so it should be fine, as I changed the location bar to work with that.
- this only effects the location bar search. If everyone is happy with this pull, I can hunt down the others and change them to use the user setting instead the hardcoded duckduckgo. I think there is a search option in some of the right click context menus.

Please let me know if you have any feedback!


app/bg/dbs/settings.js
```diff
@@ -39,7 +40,14 @@ export const setup = function (opts) {
...
     dat_bandwidth_limit_up: 0,
-    dat_bandwidth_limit_down: 0
+    dat_bandwidth_limit_down: 0,
+    selected_search_engine: 0,
+    custom_search_engine_name: '',
+    custom_search_engine_url: '',
+    default_search_engines: [
+        {name:'DuckDuckGo',url:'https://www.duckduckgo.com/'},
+        {name:'Google',url:'https://www.google.com/search'}
+      ]
```
pfrazee, line 50: I think we can modify these settings to be a little more robust. Rather than making the custom search engine separate, we can just have it be a part of the list. So:

```
selected_search_engine: 0,
search_engines: [
  {name: 'DuckDuckGo', url: 'https://www.duckduckgo.com/'},
  {name: 'Google', url: 'https://www.google.com/search'}
]
```

Then the user adds a custom engine by adding to that list. A few notes about this:

 1. The settings-db only saves strings and numbers, so we'll need to update the get/set to parse/stringify as JSON for the `search_engines` value. You can just hard code that as a conditional.
 2. Check your tab spacing there, the default_search_engines array is too indented.
 3. Also please put spaces after commas and colons in objects!

app/fg/location-bar/index.js
```diff
@@ -290,16 +289,25 @@ class LocationBar extends LitElement {
     var queryId = ++this.queryIdCounter
     this.inputValue = this.inputValue.trim()
     var finalResults
+    var settings = await bg.beakerBrowser.getSettings()
+    var searchEngine
+    //get search engine to use
+    if (settings.selected_search_engine == -1) {
```
pfrazee, line 295: The -1 logic here won't be needed after the custom search-engine gets stored to the `search_engines` array, but it _would_ be smart to have DDG as a builtin backup in case the value is bad for some reason.

app/userland/settings/js/views/general.js
```diff
@@ -59,10 +58,64 @@ class GeneralSettingsView extends LitElement {
...
+                          <input type="radio"
+                                 id="engine${i}"
+                                 name="search-engine"
+                                 value="${i}"
+                                 ?checked="${this.settings.selected_search_engine === i}"
+                                 @change="${this.onSearchEngineChange}"/>
+                          <label for="engine${i}">
+                            ${engine.name} (${engine.url})
+                          </label>
+                        </div>`
```
pfrazee, line 85: Tab depth is a little off on this. I'm okay with how you aligned the attributes but the markup chunks should be only 2 spaces deeper than the `return` statement

matthewauld:
Works for me! Will make changes.

app/bg/dbs/settings.js
matthewauld: Ahh, just store the whole JSON string. Right, that is way easier.

matthewauld:
Alright, updated - stored as a JSON string so user can add as many custom search engines as they want. Fixed the tabbing issues (I think, let me know if I didn't).  Also added logic so they can' t remove the last/current search engine. Custom search engine also affects the context menu (right-click).  Made small changes to css so that `input[type="url"]` has the same styling as `input[type="text"]`

Thanks again for feedback! This is my first contribution to an open source project, so apologies if I miss something.

app/bg/dbs/settings.js
```diff
@@ -44,7 +45,12 @@ export const setup = async function (opts) {
     default_dat_ignore: '.git\n.dat\nnode_modules\n*.log\n**/.DS_Store\nThumbs.db\n',
     analytics_enabled: 1,
     dat_bandwidth_limit_up: 0,
-    dat_bandwidth_limit_down: 0
+    dat_bandwidth_limit_down: 0,
+    selected_search_engine: 0,
+    default_search_engines: [
```
pfrazee, line 50: We should probably call this `search_engines` instead of `default_search_engines` since it's user-modifiable

app/bg/dbs/settings.js
```diff
@@ -57,10 +63,12 @@ export const once = events.once.bind(events)
 
 /**
  * @param {string} key
- * @param {string | number} value
+ * @param {string | number | object} value
  * @returns {Promise<void>}
  */
 export async function set (key, value) {
+  if (['default_search_engines', 'selected_search_engine'].includes(key))
```
pfrazee, line 70: Does `selected_search_engine` need to be JSONified? It's a number, right? Only objects need to be serialized.

pfrazee, line 70: On conditionals, please use brackets unless it's all one line, so either:

```
if (cond) dothing() // <-- only do this rarely too
```

or

```
if (cond) {
  doThings()
}
```

but never

```
if (cond)
  doThing()
```

We follow this rule to avoid accidental dangling ifs like Heartbleed suffered from.

pfrazee, line 70: It's a good idea to move the list of keys that need to be JSONified to a constant at the top of the file, something like `JSON_ENCODED_SETTINGS` so that the lists dont fall out of sync in the various parts of the code

app/bg/dbs/settings.js
```diff
@@ -107,9 +119,12 @@ export const getAll = function () {
   return setupPromise.then(v => cbPromise(cb => {
     db.all(`SELECT key, value FROM settings`, (err, rows) => {
       if (err) { return cb(err) }
-
       var obj = {}
       rows.forEach(row => { obj[row.key] = row.value })
+      // parse non-string values
+      if (obj.default_search_engines) obj.default_search_engines = JSON.parse(obj.default_search_engines)
+      if (obj.selected_search_engine) obj.selected_search_engine = JSON.parse(obj.selected_search_engine)
```
pfrazee, line 126: Make sure to use the `JSON_ENCODED_SETTINGS.includes(key)` logic here instead of hardcoding them.

app/bg/dbs/settings.js
```diff
@@ -93,7 +101,11 @@ export const get = function (key) {
   // stored values
   return setupPromise.then(() => cbPromise(cb => {
     db.get(`SELECT value FROM settings WHERE key = ?`, [key], (err, row) => {
-      if (row) { row = row.value }
+      if (row) {
+        row = row.value
+        if (['default_search_engines', 'selected_search_engine'].includes(key))
+          row = JSON.parse(row)
```
pfrazee, line 107: We should wrap the `parse()` calls in a try/catch and then fall back to the default value on failure

app/bg/dbs/settings.js
pfrazee, line 126: Same story, wrap in try/catch

app/fg/location-bar/index.js
```diff
@@ -290,16 +289,21 @@ class LocationBar extends LitElement {
     var queryId = ++this.queryIdCounter
     this.inputValue = this.inputValue.trim()
     var finalResults
+    var settings = await bg.beakerBrowser.getSettings()
```
pfrazee, line 292: We dont want to call `getSettings()` here because it gets called multiple times while the user types and needs to be quick. It should happen in `reset()` and the settings can be stored on the `this`

app/userland/settings/js/views/general.js
```diff
@@ -484,6 +545,35 @@ class GeneralSettingsView extends LitElement {
...
+  onSearchEngineChange (e) {
+    const newValue = e.target.value
+    this.settings.selected_search_engine = parseInt(newValue)
+    beaker.browser.setSetting('selected_search_engine', this.settings.selected_search_engine)
+    this.requestUpdate()
+  }
+
+  removeSearchEngine (i) {
+    // decrement selected search engine so it points to the same one if the removed index is less than current one
+    if (i<this.settings.selected_search_engine){
```
pfrazee, line 557: Sorry to nitpick but can you put spaces around the `<`?

app/userland/settings/js/views/general.js
```diff
@@ -484,6 +545,35 @@ class GeneralSettingsView extends LitElement {
     this.requestUpdate()
   }
 
+  onSearchEngineChange (e) {
+    const newValue = e.target.value
+    this.settings.selected_search_engine = parseInt(newValue)
+    beaker.browser.setSetting('selected_search_engine', this.settings.selected_search_engine)
+    this.requestUpdate()
```
pfrazee, line 552: Can you render a "Settings Updated" toast here? (See the other update handlers for how)

matthewauld: Sure!

app/userland/settings/js/views/general.js
matthewauld: No worries - thought I got them all.

app/bg/dbs/settings.js
matthewauld: Makes total sense.

app/fg/location-bar/index.js
matthewauld: Sounds good. Since `reset` is not async and `getSettings` is, I changed it so reset calls `this.settingsPromise = bg.beakerBrowser.getSettings()` and `queryAutocomplete` calls `var settings = await this.settingsPromise`. It works, but I am not sure if there is a better/proper way?

app/bg/dbs/settings.js
matthewauld: Right, Sqlite allows different datatypes in a column... I thought they would all have to be/would be converted to strings.

matthewauld:
Alright - besides being unsure of where I called getSettings in the location-bar (It works, just not sure if it is the right form) otherwise, should be good to go! Sorry for all the back and forth, appreciate the notes.

pfrazee:
Okay great work! Testing now

pfrazee:
Okay merged! Appreciate you working with me through it. I made a couple of final tweaks before merging (https://github.com/beakerbrowser/beaker/commit/d08989861200bfee44799166e825cc7de5fac585)

matthewauld:
Ah - yeah that does work better!

PR #1746, opened 2020-06-17
added missing s to searchEngine

matthewauld:
Super tiny typo in context-menu that throws error when no default search engine is selected. Is just a fallback and will probably never run, but you never know.


pfrazee:
Thanks!

Issue #1665, opened 2020-05-24
Run daemon at start-up


da2x:
**Is your feature request related to a problem? Please describe.**
Increase peer participation and hyperdrive availability.

**Describe the solution you'd like**
Start the daemon on start-up and let it run indefinitely without starting Beaker.

Pause the daemon in any of these conditions:
* Beaker isn’t running and the device is powered from battery.
* Beaker isn’t running and the device’s is on a metered network connection.

(There are APIs for both of these on Linux, macOS, and Windows.)

**Describe alternatives you've considered**
The status-quo.

matthewauld:
I can take a crack at this if it would be helpful. Basically add a setting that if enabled would launch the PM2 with noPM2DaemonMode set to false, and if disabled, kill the running hyperdrive-daemon and restart it with noPM2DaemonMode set to true? We can display a confirmation window, as this would need to be a "restart beaker to on change" type request.

The pausing on battery and metered service is a neat idea, but I think it would need to be implemented in the hyperdrive-daemon first, as it would need to operate when beaker is closed, and would effect all other clients using the daemon.

pfrazee:
@matthewauld +1 for you to hit this but it actually would require a different approach - you need to start beaker, not something else.

It looks like for windows and mac we can use a builtin API (https://www.electronjs.org/docs/api/app#appsetloginitemsettingssettings-macos-windows). For linux we may need to use https://www.npmjs.com/package/auto-launch

matthewauld:
Ah, OK - so you want beaker to keep the daemon running - sounds good. I can take a look.

pfrazee:
Yeah it's somewhat confusing but Beaker actively manages hyperdrive-daemon and ties its lifecycle together. When you close Beaker to the tray, Beaker stays running.

matthewauld:
So, just so I know I am on the right track with this: I would need to expose the 
`setBeakerAtStartup(bool)` function  (or whatever I call it) through the `browser` object in `browser.js`, so that when the user toggles the setting the change is made, right?
And I am guessing also we would check to ensure that autostart is set correctly at each startup?

pfrazee:
@matthewauld that sounds right, since it's more than just a setting in the internal settings db, putting a method on the `browser` internal RPC should be the way to go

matthewauld:
@pfrazee sorry this has taken so long - got sidetracked this week. 
I have it working on linux and will test on mac a little later today. I don't have a windows box
 
[...]

PR #1753, opened 2020-06-25
added launch at startup functionality

matthewauld:
Implements #1665 
Alright! So, this is not fully tested but:
 - works on mac
 - don't have a windows box to test, but I followed all the instructions to ensure it works with the auto-updater.
 - doesn't work when running from source (but I don't think you want it to?)
- linux works, but I cannot seem to find the correct executable path.  both process.execPath and app.getPath('exe') give you the file stored in /tmp/.mount_Beakerxxx which seems to have something to do with the builder? I did a bunch of searching, but cannot find the solution. I am sure it is straightforward and if it is easy, you just need to swap out process.execPath on line 889 of browser.js

If you need changes, just let me know!


pfrazee:
Thanks @matthewauld! I'll need to test this out when I'm close to the next build and testing on platforms.

PR #1759, opened 2020-06-30
fix onreadable edgecase

mafintosh:
fixes an edge case where if the stream emits readable before electron gets a chance to attach the listener, it deadlocks


pfrazee:
Thx thx

PR #1772, opened 2020-07-12
Fixed overlapping UI in the history tab

JonathanLindroth:
I noticed that the browsing history tab had overlapping elements when opening DevTools or using split windows on my 1366x768 monitor. 

**Before:**

I fixed it by adding a media query which sets a left margin when reaching the breakpoint. I then adjusted the width of the navbar slightly to remove the remaining overlap.

**After:**


pfrazee:
❤️ Much appreciated!

PR #1785, opened 2020-07-22
Improve mime type detection for hyper://

voxpelli:
Currently the mime type is purely detected based on file name. This is a challenge if the file name does not include the actual file type.

Speaking with @mafintosh and @digli me together with @digli concluded that we could do two things:

1. Actually read a chunk and use that to detect mime type
2. Save the mime type to the metadata and use that explicit mime type rather than implicitly derive it from the file

This PR contains both approaches – one commit for each.

### DIFF
--- a/app/bg/protocols/hyper.js
+++ b/app/bg/protocols/hyper.js
@@ -370,7 +370,11 @@ export const protocolHandler = async function (request, respond) {
       })
     }
 
-    var mimeType = mime.identify(entry.path)
+    var chunk;
+    for await (const part of checkoutFS.session.drive.createReadStream(entry.path, { start: 0, length: 512 })) {
+      chunk = chunk ? Buffer.concat(chunk, part) : part;
+    }
+    var mimeType = mime.identify(entry.path, chunk)
     if (!canExecuteHTML && mimeType.includes('text/html')) {
       mimeType = 'text/plain'
     }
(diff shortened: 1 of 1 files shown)


voxpelli:
Should be noted: I have not actually tested this code as I have not (yet) looked into how to build Beaker browser locally

pfrazee:
I actually have some code for piping the stream and identifying without having to read a throwaway chunk at https://github.com/beakerbrowser/beaker/blob/master/app/bg/lib/mime.js#L60. @mafintosh do you see any problems with using that? (I believe I stopped using it due to electron's streams being broken)

mafintosh:
@pfrazee only that, that fails if the stream is empty

voxpelli:
@pfrazee Not possible to detect as part of a pipe as it needs to be set in the headers before any streaming starts to happen?

What one could do is to not throw away the chunk, but rather reuse any part that matches the requested range and then append the rest of the stream into that one. I didn't include that here as that feels like an optimization that can be added once the approach has been agreed upon.

@pfrazee @mafintosh Does the metadata part look good? Explicit like types would be even neater than improved sniffing

pfrazee:
Ah it does also fail if a range query is happening. I'll just roll with this, the cost is minimal and the odds of this case occurring are low.

PR #1801, opened 2020-08-22
Implement ad blocking with filter list selection

Alamantus:
Adds @cliqz/adblocker to check each `webRequest` against selected ad filter lists. [EasyList](https://easylist.to/easylist/easylist.txt) is activated by default and others default lists (from the [EasyList homepage](https://easylist.to/)) are available to activate ([EasyPrivacy](https://easylist.to/easylist/easyprivacy.txt), [EasyList Cookie List](https://easylist-downloads.adblockplus.org/easylist-cookie.txt), [Fanboy's Social Blocking List](https://easylist.to/easylist/fanboy-social.txt), [Fanboy's Annoyance List](https://easylist.to/easylist/fanboy-annoyance.txt), and [Adblock Warning Removal List](https://easylist-downloads.adblockplus.org/antiadblockfilters.txt)), plus more can be added from the Settings page.

Also adds a new tab to the Settings page where the filter lists can be activated/deactivated, removed, or added.

### DIFF
--- a/app/bg/adblocker.js
+++ b/app/bg/adblocker.js
@@ -1,13 +1,20 @@
 import { session } from 'electron'
 import { FiltersEngine, Request } from '@cliqz/adblocker'
 import fetch from 'cross-fetch';
+import * as settingsDb from '../bg/dbs/settings';
 
 // exported API
 
 export async function setup () {
-  const blocker = await FiltersEngine.fromLists(fetch, [
-    'https://easylist.to/easylist/easylist.txt'
-  ])
+  const adblockLists = await settingsDb.get('adblock_lists')
+  const activeLists = adblockLists.filter(list => list.selected)
+
+  if (activeLists.length < 1) {
+    session.defaultSession.webRequest.onBeforeRequest({ urls: ['*://*/*'] }, (details, callback) => callback({ cancel: false }))
+    return
+  }
+
+  const blocker = await FiltersEngine.fromLists(fetch, activeLists.map(list => list.url))
 
   const beakerUrls = /^(beaker|blob)/
   session.defaultSession.webRequest.onBeforeRequest({ urls: ['*://*/*'] }, (details, callback) => {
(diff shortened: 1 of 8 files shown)


remusao:
Hey @Alamantus,

I had opened [another PR](https://github.com/beakerbrowser/beaker/pull/1453) to integrate `@cliqz/adblocker`, not sure if you already saw it. Since it's out-dated, I'm happy to close it in favor of yours!

In any case, did you maybe consider using `@cliqz/adblocker-electron`, which might integrate better with Beaker?

Best,

Alamantus:
Hi @remusao I did see that PR, but only after I created this one (oops). I did add an extra feature to this PR (the ability to choose filter lists from Settings), but I think the default lists that you chose are better than the ones I chose—I just used the lists on the home page of the EasyList site. Either way, whatever solution ends up being picked, I'm just excited for the functionality within Beaker!

As for `@cliqz/adblocker-electron`, I tried using it, and it didn't seem to do anything when I tested it with Beaker. I suspected that there was something else going on with Beaker's requests at this point that was causing it not to work properly, so I decided to use the `@cliqz/adblocker` directly instead. Aside from it actually working when I tested it, I think this method is beneficial because it can potentially cache filters in the future so they don't need to be fetched every time Beaker loads. (Maybe `adblocker-electron` can also do that, but I still couldn't get it to work.)

Thanks!

remusao:
Nice!

> Maybe adblocker-electron can also do that, but I still couldn't get it to work.

Yes `ElectronBlock` is a super-set of the base `Blocker` from `@cliqz/adblocker` so you can also do the caching. One advantage of the `@cliqz/adblocker-electron` is that it can leverage some features of Electron to provide better blocking on some sites (sometimes the `onBeforeRequest` listener is not enough). One possible drawback is that depending on how Beaker is packaged, it might cause issues with the preload script required for some features (but it could probably be fixed). We might also want to check if Beaker already registers webRequest listeners because we can only have one and they might conflict (this can also be fixed).

Anyway, happy to provide further feedback if this is something the Beaker people are interested in merging in!

Thanks for putting in the work!

pfrazee:
Great stuff, merged into the `search` branch which contains the next release.

Thank you @Alamantus and @remusao for your work on this. I'm sure a number of people will appreciate this.

PR #1936, opened 2021-08-12
app, scripts: bump package.json for m1 support + sqlite

matildepark:
- Bumps sqlite3 to 5.0.0 because the only breaking change is dropping support for Node <10 and knex is breaking on the sqlite4 peer dependency in the root.
- Bumps electron-builder to 22.10.4+, which is when M1 support was introduced. I expected more work here, but it does build without any further changes to the build config, and it does seem to run fine on my M1 Macbook Air?

image

Cool, I guess. This doesn't solve the problem of "needing an M1 machine to build arm64 binaries", though, as far as I know.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -87,7 +87,7 @@
     "slugify": "^1.3.5",
     "speedometer": "^1.0.0",
     "split2": "^2.2.0",
-    "sqlite3": "^4.1.0",
+    "sqlite3": "^5.0.0",
     "stream-throttle": "^0.1.3",
     "streamx": "^2.6.0",
     "textextensions": "^2.5.0",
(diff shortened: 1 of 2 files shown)


miratcan:
Is this project still active? no response to your PR since august.

Kyllingene:
No, I believe the creator has moved onto another browser project; something to do with IPFS, I believe.

RangerMauve:
Paul has moved on to working on [atek-cloud](https://github.com/atek-cloud).

I'm working on another p2p browser over [here](https://agregore.mauve.moe/) if folks want to do something more active. 😅 It's a lot more minimal than Paul's thing though.

